PR #2, opened 2018-11-13
Feature/hook up storyblok

palmfjord:
A first iteration at least. I started with Graphql and `apollo-rest-link`, but rejected that solution because this is such a simple task to fetch some arbitrary data from one rest endpoint. I see no reason for pulling in the complexity (and bundle size) of apollo and all the graphql tooling. Thoughts?


src/server/utils/storyblok.ts
```diff
@@ -0,0 +1,49 @@
+import axios from 'axios'
```
hedvigoscar: Not a big thing but do we need `axios` here? I'm not exactly against it but I'm not seeing anything that we couldn't easily accomplish with `fetch` 🙂

palmfjord: well it happens server-side so we need a library for it. I personally i like axios more than `node-fetch`, but i can switch it out if you want me to? :)

hedvigoscar: No, I think this is all good 🙂We may want to switch on `giraffe` as well given that it uses `node-fetch` 🙂

palmfjord: hm, maybe. axios' typings are quite awesome right? Unsure how it compares to `node-fetch` though?

palmfjord: Haha 😄

PR #7, opened 2018-11-13
Feature/first sections

palmfjord:
Huston - we have contact 😱

### DIFF
--- a/src/blocks/SingleQuoteBlock.tsx
+++ b/src/blocks/SingleQuoteBlock.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react'
+import { BaseBlock } from './BaseBlock'
+
+export interface SingleQuoteSectionProps extends BaseBlock {
+  quote: string
+  author: string
+  textPosition: 'left' | 'right' | 'center'
+  backgroundImage?: string
+}
+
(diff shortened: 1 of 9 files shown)


src/blocks/SingleQuoteBlock.tsx
```diff
@@ -0,0 +1,20 @@
...
+  quote: string
+  author: string
+  textPosition: 'left' | 'right' | 'center'
+  backgroundImage?: string
+}
+
+export const SingleQuoteSection: React.FunctionComponent<
+  SingleQuoteSectionProps
+> = (props) => (
+  <div>
```
isaacSennerholt: div vs section for these larger blocks. What do you think?

src/blocks/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const getBlockComponent = <
+  TBlockType extends keyof typeof blockComponents
+>(
+  blockType: string | TBlockType,
+):
+  | (typeof blockComponents)[TBlockType]
+  | React.ComponentType<BaseBlock>
+  | undefined => {
+  return blockComponents[blockType as keyof typeof blockComponents]
+}
```
isaacSennerholt: Why so any checks for type of block? Can a block type be undefined?

PR #19, opened 2018-11-19
Snake case blocks

palmfjord:
🐫 -> 🐍

ping @isaacSennerholt

Will prolly introduce a few bugs but what do you do 🤷‍♀️ lets fix them when we build each section. Sounds reasonable? :)


palmfjord:
alright @hedvigoscar done 😄

PR #21, opened 2018-11-19
Feature: accordion section

palmfjord:
screenshot 2018-11-19 at 12 43 57

What do you say @isaacSennerholt, should we add the color variations directly?


isaacSennerholt:
@palmfjord What are the cons of using colour variations right away?

palmfjord:
@isaacSennerholt i guess it's just that i wanna keep the PR size down 🤷‍♀️ not sure how to implement the color thing exactly. Maybe we can merge this guy and i'll jump on switching colors in a separate pr?

palmfjord:
@hedvigoscar @isaacSennerholt would you mind taking a second look btw? 🙈 i found some improvements after lunch break haha

isaacSennerholt:
I think a "larger" PR size is fine as long as it isn't a bunch of complicated logic. But yea, i'm fine with you merging this one and you can do it in a separate :)

PR #23, opened 2018-11-19
Safely sync storyblok on release

palmfjord:
it sucks and its error prone to do this manually. Lets automate it. This should be safe enough. Tested by hijacking staging. Seemed to have worked out fine 👌

### DIFF
--- a/bin/heroku-release
+++ b/bin/heroku-release
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+current_dir=$(dirname "$0")
+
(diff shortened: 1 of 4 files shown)


bin/heroku-release
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+space_id="${STORYBLOK_SPACE_ID:-}"
+
+if [[ -z "${space_id}" ]]; then
+    echo '$STORYBLOK_SPACE_ID not set, skipping Storyblok sync'
+    exit 0
+fi
+
+echo "running fetch-space on space id $space_id"
```
palmfjord: sync-space*

PR #24, opened 2018-11-19
Actually utilize storyblok's cache & handle invalidation

palmfjord:
When playing around with storyblok on a private project this weekend i realized they cache things _very_ hard unless you pass this timestamp. This is a proposal of how to solve the super-hard caching they do, but still utilise it.

Storing the cv in-memory could cause inconsistencies with the load balancer, but i think we can live with that? In the worst possible scenarion the end user would se inconsistent content for a few minutes if moving back and forth between pages. Maybe if we implement some caching on our side we'll utilise some redis or something to do this properly, but i'm fine with this as of now. What do you say?

### DIFF
--- a/src/server/utils/storyblok.ts
+++ b/src/server/utils/storyblok.ts
@@ -1,6 +1,15 @@
 import axios from 'axios'
 import { BodyStory } from '../../storyblok/StoryContainer'
 import { config } from '../config'
+import { appLogger } from '../logging'
+
+let remoteCacheVersion = new Date()
+const getRemoteCacheVersionTimestamp = () =>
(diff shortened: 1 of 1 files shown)


isaacSennerholt:
@palmfjord I don't think content inconsistency is a huge problem if it's just for a few minutes after a switch. I think it's a problem that we can solve if/ and then when it becomes a pain.

palmfjord:
@isaacSennerholt yeh it sounds reasonable. If we're reeeeeally unlucky the longest period of inconsistency should be 15 minutes. i think that's highly unlikely (but possible)

PR #28, opened 2018-11-20
Feature/new blocks

isaacSennerholt:
Blocks created for desktop:
• ImageTextBlock
• TitleCtaBlock
• TitleParagraphBlock

Ps. Sorry for the mess commit/PR wise :)


package-lock.json
```diff
@@ -0,0 +1,13072 @@
+{
```
palmfjord: We try to use `yarn` here, so no `package-lock`needed :) this file can be removed

package.json
```diff
@@ -33,6 +33,7 @@
     "@types/react-hot-loader": "^4.1.0",
     "@types/react-router": "^4.4.1",
     "@types/react-router-dom": "^4.3.1",
+    "@types/stack-trace": "0.0.29",
```
palmfjord: i _think_ this will solve why https://github.com/HedvigInsurance/web-next/pull/18 is failing. nice!

src/blocks/ImageTextBlock.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+const FilledButtonWithMarginComponent = styled(FilledButtonComponent)({
+  marginTop: '47px',
+})
+
+const OutlinedButtonWithMarginComponent = styled(OutlinedButtonComponent)({
+  marginTop: '47px',
+})
+
+const SectionComponent = styled('section')(
```
palmfjord: i think we should reuse https://github.com/HedvigInsurance/web-next/blob/master/src/components/blockHelpers.tsx#L3 rather than creating new ones for each block

src/blocks/ImageTextBlock.tsx
```diff
@@ -0,0 +1,123 @@
...
+  NativeColorPickerComponent,
+} from './BaseBlockProps'
+
+import {
+  FilledButtonComponent,
+  OutlinedButtonComponent,
+} from '../components/Buttons'
+
+const FilledButtonWithMarginComponent = styled(FilledButtonComponent)({
+  marginTop: '47px',
```
palmfjord: imo we should use `rem`s instead of pixels. Thoughts? :) also i think we should avoid these types of "magic numbers" when possible :)

palmfjord: i.e. `47px` -> `3rem` :) since 47/16 = ish 3

src/blocks/ImageTextBlock.tsx
palmfjord: also imo we should merge https://github.com/HedvigInsurance/web-next/pull/22 and reuse that. Thoughts?

src/blocks/ImageTextBlock.tsx
```diff
@@ -0,0 +1,123 @@
...
+  paragraph_color,
+  text_position,
+  button_title,
+  button_type,
+  show_button,
+  image,
+  background_color,
+}) => {
+  const buttonComponents = {
+    filled: FilledButtonWithMarginComponent,
```
palmfjord: this could be moved outside of the render function, so we don't have to create it on every render. We can save a few nanoseconds or something 🐌

src/blocks/ImageTextBlock.tsx
```diff
@@ -0,0 +1,123 @@
...
+  paragraph_color: NativeColorPickerComponent
+  text_position: string
+  button_title: string
+  button_type: 'filled' | 'outlined'
+  show_button: true | false
+  image: string
+  background_color: NativeColorPickerComponent
+}
+
+export const ImageTextBlock: React.SFC<ImageTextBlockInterface> = ({
```
palmfjord: react.sfc is derprecated in react 16.6, use `React.FunctionComponent` instead :)

src/blocks/TitleCtaBlock.tsx
```diff
@@ -0,0 +1,69 @@
...
+import styled from 'react-emotion'
+import { MaxWidthContainerComponent } from '../components/blockHelpers'
+import { BaseBlockProps, NativeColorPickerComponent } from './BaseBlockProps'
+
+import {
+  FilledButtonComponent,
+  OutlinedButtonComponent,
+} from '../components/Buttons'
+
+const SectionComponent = styled('section')(
```
palmfjord: same, we can reuse this

src/blocks/TitleCtaBlock.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const ContentContainer = styled('div')({
+  display: 'flex',
+  flexDirection: 'row',
+  justifyContent: 'space-between',
+  alignItems: 'center',
+  width: '100%',
+})
+
+const TitleComponent = styled('h2')(
```
palmfjord: duplicate code :D

src/blocks/ImageTextBlock.tsx
```diff
@@ -0,0 +1,123 @@
...
+            <TitleComponent titleColor={title_color.color}>
+              {title}
+            </TitleComponent>
+            <ParagraphComponent
+              paragraphColor={paragraph_color.color}
+              dangerouslySetInnerHTML={{
+                __html: paragraph.html,
+              }}
+            />
+            {show_button && <ButtonComponent>{button_title}</ButtonComponent>}
```
palmfjord: i'm pretty sure these cta's will be links rather than buttons, so we should add a target and also make them `a` tags instead of `button`s

src/blocks/TitleParagraphBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }: {
+    paragraphColor: string
+    textPosition: string
+  }) => ({
+    display: textPosition === 'right' ? 'none' : 'block',
+    marginTop: textPosition === 'center' ? '25px' : 0,
+    maxWidth: textPosition === 'left' ? '40%' : '100%',
+    width: '100%',
+    color: paragraphColor,
+    fontSize: '20px',
```
palmfjord: is it possible to avoid hard-coded font sizes? It feels weird to me that one specific paragraph should be `4px` larger than the rest of the paragraphs. Haven't looked at the design though but that's a gut feel

src/components/blockHelpers.tsx
```diff
@@ -14,3 +14,18 @@ export const ContentWrapper = styled('div')({
     padding: '0 1rem',
   },
 })
+
+export const MaxWidthContainerComponent = styled('div')({
```
palmfjord: isn't this _almost_ the same as `ContentWrapper` 10 lines above? :) any particular reason for not reusing that?

src/pages/StoryPage.tsx
```diff
@@ -3,7 +3,7 @@ import Helmet from 'react-helmet-async'
 import { getBlockComponent } from '../blocks'
 import { StoryContainer } from '../storyblok/StoryContainer'
 
-export const StoryPage: React.FunctionComponent = () => (
+export const StoryPage: React.SFC = () => (
```
palmfjord: Any particular reason for this? SFC is deprecated

src/server/page.tsx
```diff
@@ -107,6 +108,7 @@ export const getPageMiddleware: Koa.Middleware = async (ctx) => {
 
   if (!story) {
     ctx.status = 404
+    appLogger.info(`Request returned 404 due to missing storyblok story`)
```
palmfjord: NICE with logging here! However this should prolly be the request logger rather than the `appLogger`. The difference is basically that it prefixes with a request id, so you can track all log posts from a specific request through all the systems (i.e. it intercepts the `X-Request-Id` header from heroku).  example: https://github.com/HedvigInsurance/web-next/blob/8bb147b119f845128d68962b7d672d3236d68d02/src/server/page.tsx#L77

src/storyblok/StoryContainer.tsx
```diff
@@ -25,7 +25,7 @@ export interface StoryContainerProps {
   children: (props: WithStory) => React.ReactNode
 }
 
-export const StoryContainer: React.FunctionComponent<StoryContainerProps> = ({
+export const StoryContainer: React.SFC<StoryContainerProps> = ({
```
palmfjord: ?

package-lock.json
isaacSennerholt: haha lol, old habit from me as i'm generally a npm user. Will do :)

palmfjord: hehe, know the feel :D

src/blocks/ImageTextBlock.tsx
isaacSennerholt: Good points, will do

src/blocks/ImageTextBlock.tsx
isaacSennerholt: Good points, will do

src/blocks/TitleParagraphBlock.tsx
isaacSennerholt: I agree! Folled the design as is, but there should definitely be some common styles for things such as fonts, spacings etc. Can add it in another PR.

src/pages/StoryPage.tsx
isaacSennerholt: Ah, my bad

src/storyblok/StoryContainer.tsx
isaacSennerholt: Switched back to type FunctionalComponent

src/pages/StoryPage.tsx
palmfjord: so easy to miss haha 😄 i just knew because my editor screams at me because it's annotated with `@deprecated`

PR #34, opened 2018-11-21
Feature/image multi text block

palmfjord:

### DIFF
--- a/src/blocks/ImageMultiTextBlock.tsx
+++ b/src/blocks/ImageMultiTextBlock.tsx
@@ -1,7 +1,13 @@
 import * as React from 'react'
 import styled from 'react-emotion'
-import { ContentWrapper, SectionWrapper } from '../components/blockHelpers'
+import {
+  ContentWrapper,
+  MOBILE_BP_DOWN,
+  SectionWrapper,
+} from '../components/blockHelpers'
 import { BaseBlockProps, MarkdownHtmlComponent } from './BaseBlockProps'
+
+const TABLET_BP_DOWN = '@media (max-width: 800px)'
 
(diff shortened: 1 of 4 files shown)


src/blocks/ImageMultiTextBlock.tsx
```diff
@@ -0,0 +1,59 @@
...
+  <SectionWrapper color={color && color.color} size="sm">
+    <ContentWrapper>
+      <Wrapper imagePosition={image_position}>
+        <Col pad={image_position === 'left' ? 'right' : 'left'}>
+          <img src={image} />
+        </Col>
+        <Col pad={image_position === 'left' ? 'left' : 'right'}>
+          {text_items.map((item) => (
+            <div key={item._uid}>
+              <h3>{item.title}</h3>
```
isaacSennerholt: h3 is a suitable header type, though it seems to be larger in font-size here than in the Figma file. Now it's ~31.984px (base 16px) while in the Figma it's 20px.

palmfjord: woops, nice catch!

src/blocks/ImageMultiTextBlock.tsx
```diff
@@ -28,6 +28,16 @@ const Col = styled('div')(({ pad }: { pad: 'left' | 'right' }) => ({
   },
 }))
 
+const Title = styled('h3')({
+  fontSize: 20,
```
isaacSennerholt: rem?

palmfjord: Yes! I was lazy, no good 🙅‍♀️

PR #40, opened 2018-11-21
Feature: Bullet point block

palmfjord:
screenshot 2018-11-21 at 17 37 19

### DIFF
--- a/src/blocks/BulletPointBlock.tsx
+++ b/src/blocks/BulletPointBlock.tsx
@@ -1,0 +1,71 @@
+import * as React from 'react'
+import styled from 'react-emotion'
+import {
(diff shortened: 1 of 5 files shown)


isaacSennerholt:
Should we separate this block from? :
screen shot 2018-11-21 at 17 41 31

Was working on a generic bullet point component that could take any of a list of bullet components, but is it better to separate the different types of bullet lists? I.e "image bullet list" from "icon bullet list".

Here's the icon bullet list i'm referring to:
screen shot 2018-11-21 at 17 41 31

palmfjord:
@isaacSennerholt aah i get ya! Idk actually, they look visually very different so maybe it's better to have them separate?

PR #41, opened 2018-11-21
Make transparent header work when scrolling

palmfjord:
looked pretty crappy before 😅

ezgif-1-5c6546663560

also, i know we despise classes, but it's virtually impossible to use `ref`s efficiently without them :(

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -4,7 +4,11 @@ module.exports = {
     '@babel/preset-react',
     ['@babel/preset-typescript', { isTSX: true, allExtensions: true }],
   ],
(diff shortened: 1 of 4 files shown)


Iamsamwhoami:
Tried ’React.createRef()’? Then you might be able to avoid the class 😈

palmfjord:
@iamsamwhoami createRef seems cool but I don’t really understand how it will help get rid of the class? we shouldn’t store it in a singleton var if that’s what you mean, especially since we’re doing SSR. And since refs can’t really be handled with `state` I don’t really see any options (but I’m also on the fly so I might have missed something in the docs)
Can you give an example?

PR #69, opened 2018-11-28
Bump node and add .nvmrc

palmfjord:
🐋


.nvmrc
```diff
@@ -0,0 +1 @@
+v10.14.0
```
hedvigoscar:
```suggestion
v11.2.0
```

:wink:

palmfjord: i don't agree, i see no reason to not use an lts version in prod unless there's something new and fancy we really want.

hedvigoscar: Agreed!

PR #82, opened 2018-11-30
Feature/Blog posts

palmfjord:
Adds blog posts. <strike>Not 100% finished as we need tags too, but it's quite a refactor so feel free to start a review :)</strike>


src/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,36 @@
...
+  display: 'inline-block',
+  fontSize: '0.9rem',
+  '&:not(:first-of-type):before': {
+    content: '"\\203A"', // rsaquo
+    display: 'inline-block',
+    padding: '0 .75rem',
+  },
+})
+
+const PlainBreadcrumb = styled('span')({})
```
hedvigoscar: Is this necessary? 🙂

palmfjord: prolly not, want me to remove it? :)

hedvigoscar: Yes, ✅  after 🙂

PR #83, opened 2018-11-30
Add ability to add blog post author

palmfjord:


palmfjord:
`heroku addons:create heroku-redis` imo at least :) pulled it in because I want to utilize it for caching later

palmfjord:
@hedvigoscar heroku redis is fairly cheap considering we store so little data in there. If not, heroku redis is just a wrapper around elasticache if i remember it correctly (but i may be wrong)

src/server/utils/redis.ts
```diff
@@ -0,0 +1,8 @@
+import * as redis from 'redis'
```
hedvigoscar: I've been told somewhere that `ioredis` is a good package. We 'implicitly' use it in giraffe, could be worth taking a look at?

palmfjord: looks very neat! nice catch

palmfjord: or "catch".. idk. input?

PR #85, opened 2018-12-03
Fix/app link

palmfjord:
We should probably utilize the app link in the header


src/utils/tracking/trackEvent.ts
```diff
@@ -0,0 +1,18 @@
+export const trackEvent = (
```
hedvigoscar: I'd suggest using `quepasa` for this instead 🙂

palmfjord: sure! is it ready tho? we kinda need to get this out as we're shipping traffic to the wrong place now 🙈

hedvigoscar: You can do this for a while, but you should consider replacing it very soon (`quepasa` should be stable enough this week)

palmfjord: amaze! ping me when its ready? :D

PR #87, opened 2018-12-04
Blog

palmfjord:
#88 is merged and all content is published, lets ship this puppy!


src/server/middlewares/states.ts
```diff
@@ -0,0 +1,37 @@
+import { Middleware } from 'koa'
+import { getBlogPosts } from '../utils/storyblok'
+import { getCachedTeamtailorUsers } from '../utils/teamtailor'
+
+export const addTeamtailorUsersToState: Middleware = async (ctx, next) => {
```
hedvigoscar: I like this! If I understand correctly it means that we only load the TeamTailor users when accessing a page that requires them, right? 🙂

src/utils/meta.tsx
```diff
@@ -1,39 +1,46 @@
...
   <>
-    <title>{story.name}</title>
+    <title>{title ? title : story && story.name}</title>
     <link
       rel="canonical"
-      href={`${getPublicHost()}/${story.full_slug.replace(/\/?home$/, '')}`}
+      href={`${getPublicHost()}/${fullSlug || getFullSlugFromStory(story)}`}
     />
-    {story.content.robots && (
+    {story && story.content.robots && (
```
hedvigoscar: Doesn't this code hurt you a little? :wink: I feel like there's probably a good ramda-style function for doing this 🙂

hedvigoscar:
Ok! Will it still be served as quickly as it would be from our CDN? 🙂

hedvigoscar:
Maybe it's a setting for the CF we have in front of `web-next`? 🙂

palmfjord:
maybe there'll be a couple of milliseconds latency more, but it's still behind cloudfront and has the same caching settings so i don't think it's nociable

palmfjord:
@hedvigoscar how do you mean? :)

PR #92, opened 2018-12-04
Feature/about us page

palmfjord:
Basically copy-pasted from https://github.com/HedvigInsurance/web/tree/master/src/sections/about-us with some sugar on top to match this repo. Sorry for the big-ass PR, but most of this code is not new so i figured its better to just push everything in one go


src/blocks/HedvigersBlock/Mobile/index.tsx
```diff
@@ -0,0 +1,112 @@
+import { fonts } from '@hedviginsurance/brand'
+// @ts-ignore
```
hedvigoscar: What's up here? 🙂

src/blocks/ImageBlock.tsx
```diff
@@ -0,0 +1,77 @@
+import { colors } from '@hedviginsurance/brand/dist'
```
hedvigoscar: 🤔

src/blocks/HedvigersBlock/Mobile/index.tsx
palmfjord: no types and i was lazy haha. Will fix

src/blocks/ImageBlock.tsx
palmfjord: oops! I blame webstorm 😂

src/blocks/HedvigersBlock/Mobile/index.tsx
palmfjord: actually, diving into this. Would it be _ok_ to skip? `swipeable-views` literally has 100 lines of proptypes, defining the props for this dep will be a hassle

palmfjord: actually, diving into this. Would it be _ok_ to skip? `swipeable-views` literally has 100 lines of proptypes, defining the props for this dep will be a hassle

hedvigoscar: https://www.youtube.com/watch?v=-lI3_b_bHLY

hedvigoscar: (go ahead and keep the ignore 😉 )

PR #96, opened 2018-12-06
Feature/animated cta block

isaacSennerholt:


storyblok/components.json
```diff
@@ -1536,6 +1536,234 @@
       "all_presets": [],
       "preset_id": null,
       "real_name": "title_paragraph_item"
+    },
+    {
+      "name": "animated_phone_block",
```
palmfjord: don't we have this already? feels like a merge issue

storyblok/components.json
```diff
@@ -1536,6 +1536,234 @@
...
+      "image": null,
+      "preview_field": null,
+      "is_root": false,
+      "is_nestable": true,
+      "all_presets": [],
+      "preset_id": null,
+      "real_name": "animated_cta_block"
+    },
+    {
+      "name": "animated_cta_item",
```
palmfjord: _imo_ it's fine to hard-code 3 items, what's your thoughts?

storyblok/components.json
```diff
@@ -1536,6 +1536,234 @@
...
+            {
+              "value": "bellNotification",
+              "name": "Bell Notification"
+            }
+          ],
+          "required": true,
+          "default_value": "left",
+          "pos": 0
+        },
+        "title": {
```
palmfjord: this should be marked as translatable

storyblok/components.json
```diff
@@ -1536,6 +1536,234 @@
...
+          "type": "text",
+          "required": true,
+          "default_value": "",
+          "pos": 1
+        },
+        "paragraph": {
```
palmfjord: also translatable

src/blocks/AnimatedCtaBlock.tsx
```diff
@@ -0,0 +1,134 @@
...
+  >
+}
+
+interface AnimationMapProps {
+  progressiveSuccess: any
+  smallMobileChat: any
+  bellNotification: any
+}
+
+const animationMap: AnimationMapProps = {
```
palmfjord: this needs to be lazily loaded, otherwise the code splitting doesn't matter as the imports will start instantly when the main bundle loads rather than when this block is actually used

palmfjord: the easiest way i think is to wrap the imports in funcitons and call them when we need them:

```ts
const animationMap = {
  progressiveSuccess: () => import(...'anmiations/...'),
 // ...
}

// ...

<LazyLottie options={{ animationData: anmiationMap[bullet.animation_type]() }} />
```

src/blocks/AnimatedCtaBlock.tsx
```diff
@@ -0,0 +1,134 @@
...
+  progressiveSuccess: import(/* webpackChunkName: "progressive-success-animation" */ 'animations/progressiveSuccess.json'),
+  smallMobileChat: import(/* webpackChunkName: "small-mobile-chat-animation" */ 'animations/smallMobileChat.json'),
+  bellNotification: import(/* webpackChunkName: "bell-notification-animation" */ 'animations/bellNotification.json'),
+}
+
+export const AnimatedCtaBlock: React.FunctionComponent<
+  AnimatedCtaBlockProps
+> = ({ color, title, bullet_points }) => (
+  <SectionWrapper color={color && color.color}>
```
palmfjord: shouldn't we pass `size` here?

storyblok/components.json
palmfjord: it'll just look weird with 1 or 2 anyways right?

src/blocks/AnimatedCtaBlock.tsx
```diff
@@ -0,0 +1,134 @@
...
+import {
+  ContentWrapper,
+  CONTENT_GUTTER,
+  MOBILE_BP_DOWN,
+  SectionWrapper,
+} from '../components/blockHelpers'
+import { LazyLottie } from '../components/LazyLottie'
+import { BaseBlockProps, MarkdownHtmlComponent } from './BaseBlockProps'
+
+const TABLET_BP_DOWN = '@media (max-width: 800px)'
```
frapperino: Should this also be in blockHelpers?🤔

storyblok/components.json
isaacSennerholt: The main purpose of separating CTA's into their own blocks is to allow modification of copy from the CMS. Is there any downside with having them separate?

src/blocks/AnimatedCtaBlock.tsx
isaacSennerholt: Good point!

storyblok/components.json
palmfjord: what i mean is just to have three fields, like `title_1`, `title_2` and so on :) so the copy is still changeable from the cms, just the number of items aren't :D

isaacSennerholt: I understand, but then you need duplicates for all dynamic fields which feels kind of messy :) I don't see a negative with having them separate in hindsight, i would've considered taking your position previous to making the block due to increased workload of creating a separate block.

src/blocks/AnimatedCtaBlock.tsx
isaacSennerholt: There's no size property on the animated cta block. I don't set a size due to SectionWrapper having a suitable default size, which is 'lg'.

palmfjord: right 😅 haha my bad, must've been the friday-confusion kickin in

src/blocks/AnimatedCtaBlock.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const AnimatedCtaBody = styled('div')()
+
+const AnimatedCtaTitle = styled('h3')({
+  fontSize: '1.375rem',
+  margin: 0,
+})
+
+const AnimatedCtaParagraph = styled('div')({
+  fontSize: '1.125rem',
```
palmfjord: i'm not saying this particular one is wrong, but i'm concerned that we effectively use `16px` font sizes in most places and `18px` in _some_ places. Feels weird?

PR #97, opened 2018-12-06
Allow buttons to be branch links

palmfjord:
🛤 🌲

### DIFF
--- a/src/blocks/ImageTextBlock.tsx
+++ b/src/blocks/ImageTextBlock.tsx
@@ -16,6 +16,7 @@ import {
   getStoryblokLinkUrl,
   Image as StoryblokImage,
 } from '../utils/storyblok'
+import { AppLink } from '../components/AppLink'
 
 const ButtonLinkWithMargin = styled(ButtonLink)({
   marginTop: '1.7rem',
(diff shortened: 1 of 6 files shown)


src/blocks/ImageTextBlock.tsx
```diff
@@ -124,16 +127,31 @@ export const ImageTextBlock: React.FunctionComponent<ImageTextBlockProps> = ({
...
-            <ButtonLinkWithMargin
-              href={getStoryblokLinkUrl(button_link)}
-              type={button_type}
-              size="sm"
-              bold
-            >
-              {button_title}
-            </ButtonLinkWithMargin>
-          )}
+          {show_button &&
```
hedvigoscar: Could it be desirable to split these down to `show_regular_button` and `show_branch_button` or something similar? This condition feels a bit weird as it is now.

hedvigoscar: Or perhaps even an enum, `NO_BUTTON, REGULAR_BUTTON, BRANCH_BUTTON` 🙂

palmfjord: I  see what you mean, unsure if i agree tho. It kinda makes sense to me to do the same in all blocks.

palmfjord:
@hedvigoscar you got it 🙌

PR #98, opened 2018-12-06
Feature/icon forest block

isaacSennerholt:


src/blocks/IconForestBlock.tsx
```diff
@@ -0,0 +1,75 @@
...
+  fontSize: '3.75rem',
+  width: '100%',
+  margin: 0,
+  [TABLET_BP_DOWN]: {
+    fontSize: '2.25rem',
+  },
+})
+
+const Picture = styled('picture')({
+  maxWidth: '47.125rem',
```
palmfjord: i don't think it matters much if this is `47` or a`47.125` rems does it? I think we should avoid these "magic" decimals if we can

src/blocks/IconForestBlock.tsx
```diff
@@ -0,0 +1,75 @@
...
+  [TABLET_BP_DOWN]: {
+    maxWidth: '100%',
+    width: 'auto',
+    display: 'block',
+  },
+})
+
+const Paragraph = styled('div')({
+  display: 'block',
+  fontSize: '1.125rem',
```
palmfjord: are we sure we want `1.125rem`s here? Considering the rest of the site uses `16px` instead of `18px` font size i think it'll look weird if we make this particular section `18px`?

src/blocks/IconForestBlock.tsx
```diff
@@ -0,0 +1,75 @@
...
+  paragraph,
+  color,
+}) => {
+  return (
+    <SectionWrapper color={color && color.color}>
+      <ColumnContentWrapper>
+        <Title>{title}</Title>
+        <Picture>
+          <source
+            media={`(max-width: 47.125rem)`}
```
palmfjord: same here, do we really need `.125`?

src/components/blockHelpers.tsx
```diff
@@ -66,7 +67,7 @@ const sectionSizeStyles = {
     },
   },
   lg: {
-    padding: '7rem 0',
+    padding: '7.8125rem 0',
```
palmfjord: is this intentional?

storyblok/components.json
```diff
@@ -1536,6 +1536,268 @@
...
+      "name": "icon_forest_block",
+      "display_name": "",
+      "schema": {
+        "color": {
+          "type": "custom",
+          "field_type": "hedvig-limited-color-picker",
+          "options": [],
+          "pos": 0
+        },
+        "title": {
```
palmfjord: translatable

storyblok/components.json
```diff
@@ -1536,6 +1536,268 @@
...
+        },
+        "paragraph": {
+          "type": "custom",
+          "required": true,
+          "default_value": "",
+          "pos": 1,
+          "field_type": "markdown-html",
+          "options": [],
+          "display_name": "",
+          "translatable": true
```
palmfjord: yay! 🙌

src/blocks/IconForestBlock.tsx
isaacSennerholt: I agree :) I'll stick to integers in the future.

src/components/blockHelpers.tsx
isaacSennerholt: Yes, as there's quite a bit of difference between the two. 7rem = 112px, 7.8125rem = 125px. I always try to think about the "px" size first, and the consider what rems we want, e.g is 125px a good number or does it feel arbitrary/random? If not, then we should apply the rem that gives us 125px, no matter the number. Though a compromise can be made when taking the multiplication table into consideration, though such compromise should always stem from the perspective of a nice px, and not rem.

storyblok/components.json
isaacSennerholt: haha, i'm apparently bad at remembering the "translatable" property lol

src/blocks/IconForestBlock.tsx
palmfjord: totally agree!

src/components/blockHelpers.tsx
palmfjord: i think we should abandon rems if we "mentally" convert them into pixels anyways tho? and use media queries + px instead

src/blocks/CardChecklistBulletPointBlock.tsx
```diff
@@ -0,0 +1,143 @@
+import { colors } from '@hedviginsurance/brand'
```
palmfjord: this file dosen't look like an icon forest block 😅

PR #99, opened 2018-12-07
Feature/Download block

palmenhq:
⬇️

### DIFF
--- a/storyblok/components.json
+++ b/storyblok/components.json
@@ -431,40 +431,34 @@
       "name": "download_block",
       "display_name": null,
       "schema": {
-        "color": {
-          "type": "custom",
-          "field_type": "hedvig-limited-color-picker",
-          "options": [],
-          "pos": 0
-        },
         "title": {
           "type": "text",
           "translatable": true,
-          "pos": 1
+          "pos": 0
         },
         "cta_text": {
           "type": "text",
           "translatable": true,
-          "pos": 2
+          "pos": 1
         },
         "error_text": {
           "type": "text",
           "translatable": true,
           "default_value": "Ojdå! Det gick inte att skicka sms till det angivna numret. Dubbelkolla numret och prova igen.",
-          "pos": 3
+          "pos": 2
         },
         "success_text": {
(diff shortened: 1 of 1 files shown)


src/blocks/DownloadBlock/AppLinkForm.tsx
```diff
@@ -0,0 +1,172 @@
...
+      utmParams,
+      defaultBranchLinkOptions,
+    )
+    ;(window as any).branch.sendSMS(
+      phoneNumber,
+      {
+        ...linkOptions,
+        data: {
+          ...(this.props.linkOptions || {}),
+          $custom_sms_text: 'Ladda ner Hedvig-appen: {{ link }}',
```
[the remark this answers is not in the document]
palmfjord: Fair enough!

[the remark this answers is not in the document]
palmfjord: actually (after irl discussion) lets push it for later. It'll be hard to do nicely with the "templating" and its better to leave it than to do a semi-good solution that we "should" revisit sometime in the future

src/blocks/DownloadBlock/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+import * as React from 'react'
+import styled from 'react-emotion'
+
+import {
+  ContentWrapper,
+  MOBILE_BP_DOWN,
+  SectionWrapper,
+} from '../../components/blockHelpers'
+import { BaseBlockProps } from '../BaseBlockProps'
+import { TABLET_BP_UP } from '../HeaderBlock/mobile'
```
isaacSennerholt: Should we keep these in one place? Most of them currently exist in blockHelpers. Do we want to move them into separate files that are more descriptive?

palmfjord: woops nice catch! Haha i think i just imported it from the wrong place

palmfjord: (ide magic is dangerous sometimes)

PR #101, opened 2018-12-10
Security/own proxy + upgrades

palmfjord:
👩‍🚒


palmfjord:
@isaacSennerholt thanks! Haha i know right? wicked

PR #103, opened 2018-12-10
Feature/image text block background

isaacSennerholt:


src/components/blockHelpers.tsx
```diff
@@ -88,12 +88,20 @@ export const SectionWrapper = styled('section')(
...
     size = 'lg',
+    backgroundImage = 'none',
   }: {
     color?: colorComponentColors
     size?: SectionSize
+    backgroundImage?: string
   }) => ({
     ...getSectionSizeStyle(size),
     ...getColorStyles(color),
+    ...(backgroundImage && {
```
palmfjord: technically this'll never be falsy since we default to `'none'` right?

storyblok/components.json
```diff
@@ -904,7 +909,7 @@
           ],
           "required": true,
           "default_value": "filled",
-          "pos": 7
+          "pos": 8
```
palmfjord: maybe we should change the rest of the `"pos": N`s also?

src/blocks/ImageTextBlock.tsx
```diff
@@ -96,6 +96,7 @@ interface ImageTextBlockProps extends BaseBlockProps {
   button_link: LinkComponent
   show_button: boolean
   image: StoryblokImage
```
palmfjord: maybe this should be optional now?

PR #110, opened 2018-12-11
Add card checklish bullet point block

isaacSennerholt:


storyblok/components.json
```diff
@@ -1617,6 +1614,149 @@
...
+          "pos": 2,
+          "field_type": "markdown-html",
+          "options": [],
+          "translatable": true
+        },
+        "check_list": {
+          "type": "options",
+          "pos": 3,
+          "translatable": true,
+          "options": [
```
palmfjord: clever! However.. any particular reason for not making this a free-text field?

palmfjord: i mean this will a) never be possible to translate and b) impossible for marketing to change if they want to reuse the block for anything else

src/blocks/CardChecklistBulletPointBlock.tsx
```diff
@@ -0,0 +1,156 @@
...
+            <BulletPointBody>
+              <BulletPointTitle>{bullet.title}</BulletPointTitle>
+              <BulletPointParagraph
+                dangerouslySetInnerHTML={{
+                  __html: bullet.paragraph && bullet.paragraph.html,
+                }}
+              />
+              <div>
+                <BulletPointChecklist>
+                  {bullet.check_list.slice(0, 8).map((check) => {
```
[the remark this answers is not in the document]
isaacSennerholt: To only display 8 checkboxes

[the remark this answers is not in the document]
palmfjord: Why should we limit it to 8?

storyblok/components.json
isaacSennerholt: It's intentional that it can't be reused, it's dedicated to this particular use case. But maybe it would be better to add checklist support to image-bullet-point-item's free text instead?

src/blocks/CardChecklistBulletPointBlock.tsx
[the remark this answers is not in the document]
isaacSennerholt: The designs specifies two columns of four checkboxes in each column. Therefor i assumed we ought have a limit of 8 checkboxes in total. There has to be a limit on the height two get a two column layout using flexbox. If it's 8 or something else doesn't matter too much, it's our own preference,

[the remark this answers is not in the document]
palmfjord: alright, coolio! Had no idea you had to use a max height 🤷‍♀️ carry on 👜

src/blocks/CardChecklistBulletPointBlock.tsx
```diff
@@ -0,0 +1,156 @@
...
+                    return (
+                      <BulletPointChecklistItem>
+                        <CheckIcon>
+                          <img src="/assets-next/tick.svg" />
```
palmfjord: where did this file go?

storyblok/components.json
isaacSennerholt: Great idea! Much better to have it as separate blocks

src/blocks/CardChecklistBulletPointBlock.tsx
```diff
@@ -0,0 +1,160 @@
...
+})
+
+const CheckIcon = styled('div')({
+  display: 'inline-block',
+  padding: '0.25rem',
+  color: colors.WHITE,
+  textAlign: 'center',
+  verticalAlign: 'middle',
+  borderRadius: '50%',
+  backgroundColor: '#1BE9B6',
```
palmfjord: `colors.GREEN`? No?

src/blocks/CardChecklistBulletPointBlock.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  __html: bullet.paragraph && bullet.paragraph.html,
+                }}
+              />
+              <div>
+                <BulletPointChecklist>
+                  {bullet.check_list.map((check) => {
+                    return (
+                      <BulletPointChecklistItem>
+                        <CheckIcon>
+                          <img src="/assets-next/tick.svg" />
```
palmfjord: still think this should be jsx rather than svg + img

src/blocks/CardChecklistBulletPointBlock.tsx
```diff
@@ -139,11 +146,11 @@ export const CardChecklistBulletPointBlock: React.FunctionComponent<
               />
               <div>
                 <BulletPointChecklist>
-                  {bullet.check_list.map((check) => {
+                  {bullet.check_list.map((check, id) => {
                     return (
-                      <BulletPointChecklistItem>
+                      <BulletPointChecklistItem key={id}>
```
palmfjord: no index as key plz 🙈 can't we use `check._uid`?

isaacSennerholt: I'm getting lazy now towards the end :/

palmfjord: haha :D

storyblok/components.json
```diff
@@ -1759,6 +1756,126 @@
...
+            },
+            {
+              "value": "right",
+              "name": "Right"
+            }
+          ],
+          "default_value": "left",
+          "pos": 1
+        },
+        "enforce_size": {
```
palmfjord: i dont think we need this right?

PR #113, opened 2018-12-12
Remove size manipulation on text

isaacSennerholt:


src/blocks/ImageTextBlock.tsx
```diff
@@ -47,24 +47,23 @@ const TextWrapper = styled('div')(
...
 
-const Title = styled('h2')(
-  ({
-    size,
-    displayOrder,
-  }: SectionSizeProps & { displayOrder: 'top' | 'bottom' }) => ({
-    fontSize: size === 'xl' ? '4.5rem' : '2.5rem',
+const Title = styled('h1')(
+  ({ displayOrder }: { displayOrder: 'top' | 'bottom' }) => ({
+    fontSize: '4.5rem',
```
palmfjord: just double-checknig are we sure we always want the `xl` option here? :)

isaacSennerholt: Yes, for now :)

PR #124, opened 2018-12-20
Defer off-screen images

palmfjord:
🌄
gives us a "slight" performance increase in lighthouse
screenshot 2018-12-20 at 09 47 18


palmfjord:
@gustafguner haha YES, apparently they're more evil than we thought also.

PR #143, opened 2019-01-07
Add title to bullet point block

isaacSennerholt:


palmfjord:
idk, lets kill this for now and revisit if we miss it 🤷‍♀️

PR #163, opened 2019-02-18
Feature/deepeq and branch links

palmfjord:
- Only show components that changed when updating storyblok
- Make the branch link option translatable everywhere


palmfjord:
@Iamsamwhoami haha 😄 the first bullet point is just that before we displayed all blocks as about to be updated in the update script (that provisions storyblok), but now we only show the ones actually changed.

PR #241, opened 2019-07-04
Dimension-based language support

palmenhq:
🇬🇧🇸🇪

we need to figure out a good way to migrate content before we merge this

### DIFF
--- a/src/utils/meta.tsx
+++ b/src/utils/meta.tsx
@@ -10,7 +10,7 @@ interface Meta {
 }
 
 const getFullSlugFromStory = (story?: Story) =>
-  story && story.full_slug.replace(/\/?home$/, '')
+  story && story.full_slug.replace(/\/?home$/, '').replace(/\/sv($|\/)/, '')
 
 const getPageTitleFromStory = (story?: Story) => {
   if (!story) {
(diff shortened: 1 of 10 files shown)


src/utils/meta.tsx
```diff
@@ -10,7 +10,7 @@ interface Meta {
 }
 
 const getFullSlugFromStory = (story?: Story) =>
-  story && story.full_slug.replace(/\/?home$/, '')
+  story && story.full_slug.replace(/\/?home$/, '').replace(/\/sv($|\/)/, '')
```
palmenhq: `/^\/sv($|\/)/`?

hedvigoscar: I cant tell whether this is correct or not 🤔

palmenhq: actually, as the `sv` part is [being sanitized](https://github.com/HedvigInsurance/web-next/pull/241/files#diff-32c50dfc6f7e52967a9eaa0e836e0d0cR109) i think this might not even be needed

palmenhq:
@hedvigoscar in short, we're not using one-to-one field mapping while translating things but rather separate directories with separate pages for language "dimension"

hedvigoscar:
Sounds good :+1:

palmenhq:
@hedvigoscar can i haz another approve plz? :D

PR #249, opened 2019-07-15
Switch Headline font to Geomanist

hedvigoscar:


app.json
```diff
@@ -3,6 +3,10 @@
   "env": {
     "USE_AUTH": "true",
     "USE_HELMET": "true",
+    "SYNC_STORYBLOK_ON_RELEASE": "false",
+    "REDIS_URL": {
+      "required": true
```
palmenhq: Maybe we should spin up our own redis for each pr instance instead? So we don't mix the caches, i smell weird issues from using the same cache for multiple servers 😬

hedvigoscar: +1

hedvigoscar: Fixed 😬

hedvigoscar:
@palmenhq Whenever you have time 😬

app.json
```diff
@@ -16,7 +17,12 @@
       "required": true
     }
   },
-  "addons": [],
+  "addons": [
+    {
+      "plan": "heroku-redis",
+      "as": "REDIS"
```
palmenhq: Should we perhaps specify the plan? Preferably hobby dev or something

hedvigoscar:
@palmenhq Done :+1:

PR #250, opened 2019-07-15
Header height, tweaks and new button in the mobile header

hedvigoscar:


hedvigoscar:
Screenshot 2019-07-16 at 14 31 26

src/blocks/HeaderBlock/mobile.tsx
```diff
@@ -52,28 +55,26 @@ export const CrossBurger = styled('div')(
...
-  position: 'absolute',
   width: '100%',
   top: '50%',
   left: 0,
   right: 0,
   height: 3,
   backgroundColor: isOpen && !isClosing ? 'transparent' : 'currentColor',
   transition: 'background-color 300ms',
-  transform: 'translateY(-1.5px)',
+  transform: 'translateY(-1px)',
```
palmenhq: when i come to think of it i think this should be half of `height` which is 3 and that's why it was 1.5. any particular reason for changing it?

app.json
```diff
@@ -28,4 +34,4 @@
       "url": "heroku/nodejs"
     }
   ]
-}
+}
```
palmenhq: ):

src/blocks/HeaderBlock/mobile.tsx
hedvigoscar: This is the half-pixel thing that made it look off on different screens. I guess we could try to cover it by using half-pixels on the other places as well? But this looks correct I think 🤷‍♀

app.json
hedvigoscar: I have no idea why it did this 😞

PR #255, opened 2019-07-17
Remove sizing and inherent spacing from blocks

hedvigoscar:
Spacer Blocks will instead be used

### DIFF
--- a/src/components/blockHelpers.tsx
+++ b/src/components/blockHelpers.tsx
@@ -66,12 +66,12 @@ const colorMap = {
     background: colors.PURPLE,
   },
   'yellow-light': {
-    color: colors.LIGHT_YELLOW,
-    background: colors.OFF_BLACK_DARK,
+    color: colors.OFF_BLACK_DARK,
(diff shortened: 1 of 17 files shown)


hedvigoscar:
@palmenhq You're absolutely right. I've gone through all the blocks and removed all the margins as best as I could.

src/components/blockHelpers.tsx
```diff
@@ -114,12 +134,13 @@ export const SectionWrapper = styled('section')(
   }) => ({
     ...getSectionSizeStyle(size),
     ...getColorStyles(color),
-    ...(backgroundImage !== 'none' && {
-      backgroundImage: `url(${backgroundImage})`,
-      backgroundSize: 'cover',
-      backgroundPosition: 'center center',
-      backgroundRepeat: 'no-repeat',
-    }),
+    ...(backgroundImage !== 'none' &&
```
palmenhq: maybe we could move this logic to its own function, there's a bit of duplicate code here

PR #257, opened 2019-07-18
CTA Styling support

hedvigoscar:


src/components/buttons.tsx
```diff
@@ -1,54 +1,93 @@
...
-    backgroundColor: 'transparant',
-    color: colors.GREEN,
-  },
-  plain: {
-    border: 'none',
-    padding: '0',
-    backgroundColor: 'transparant',
-    color: colors.GREEN,
-  },
+type ColorMap = { [K in colorComponentColors]: string }
```
palmenhq: i guess also this could be something like `Record<colorComponentColors, string>` if we want :)

hedvigoscar:
I was considering doing this, but as it would contain some pretty nested fields (storyblok-links, colors etc.) I decided against it as it would be a bit too complex I think.

PR #259, opened 2019-07-18
Press Card Block

hedvigoscar:


src/blocks/PressCardBlock/PressCardItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+const PressItemContainer = styled('div')({
+  borderRadius: 8,
+  padding: 24,
+  margin: '0 20px',
+  maxWidth: 330,
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'center',
+  justifyContent: 'center',
+  backgroundColor: 'white',
```
palmenhq: should this always be white, no matter what the block background is?

src/blocks/PressCardBlock/PressCardItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+  marginTop: 5,
+  textAlign: 'center',
+  color: colors.OFF_BLACK_DARK,
+})
+
+export const PressCardItem: React.FunctionComponent<PressCardItemProps> = ({
+  logo,
+  quote,
+}) => (
+  <PressItemContainer>
```
palmenhq: `*Wrapper` plz :) container = state container

src/blocks/PressCardBlock/index.tsx
```diff
@@ -0,0 +1,71 @@
+// @ts-ignore
```
palmenhq: ): can we write some types for this, or in _some_ other way avoid it than turning off the compiler?

src/blocks/PressCardBlock/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  <SectionWrapper color={color && color.color} size="sm">
+    <Wrapper id="press_card">
+      <MediaQuery query="(max-width: 700px)">
+        <Swiper resistance enableMouseEvents slideStyle={{ display: 'flex' }}>
+          {cards.map((card) => (
+            <PressCardItem key={card._uid} {...card} />
+          ))}
+        </Swiper>
+      </MediaQuery>
+      <MediaQuery query="(min-width: 700px) and (max-width: 900px)">
```
palmenhq: technically both these sliders will be visible on 700px no? so i think this should be `701px`

src/blocks/PressCardBlock/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+          resistance
+          enableMouseEvents
+          slideStyle={{ display: 'flex', width: '50%' }}
+        >
+          {cards.map((card) => (
+            <PressCardItem key={card._uid} {...card} />
+          ))}
+        </Swiper>
+      </MediaQuery>
+      <MediaQuery query="(min-width: 900px)">
```
palmenhq: and same here, this should prolly be `901px`

hedvigoscar:
For reference, this block is mostly a clone of `PressBlock` mixed with `HedvigersBlock`. Most of the issues in the code are live in these blocks as well.

src/blocks/PressCardBlock/PressCardItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+  flexDirection: 'column',
+  alignItems: 'center',
+  justifyContent: 'center',
+  backgroundColor: 'white',
+})
+
+const Logo = styled(DeferredImage)({
+  width: 80,
+  height: 80,
+  objectFit: 'contain',
```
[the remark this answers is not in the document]
hedvigoscar: It does indeed look bad on IE11

src/blocks/PressCardBlock/PressCardItem.tsx
axelbacklund: Is off white now when block background is white

src/blocks/PressCardBlock/PressCardItem.tsx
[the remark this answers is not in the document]
axelbacklund: Fixed by setting a fixed image height and letting the width vary, keeping the objectFit however to in modern browsers manage the unlikely cases of really narrow images!

src/blocks/PressCardBlock/PressCardItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+  color: colors.OFF_BLACK_DARK,
+})
+
+export const PressCardItem: React.FunctionComponent<PressCardItemProps> = ({
+  logo,
+  quote,
+  blockColor,
+}) => (
+  <PressItemWrapper offWhite={blockColor && blockColor.color === 'standard'}>
+    <Text>“{quote}”</Text>
```
palmenhq: Imo we shouldn’t hard-code quotes, it’s very easy for the content guys to add and will be limiting in the cases they would not want it

src/blocks/PressCardBlock/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export interface PressCardItemProps {
+  _uid: string
+  quote: string
+  logo: Image
+  blockColor?: ColorComponent
+}
+
+interface PressBlockProps extends BaseBlockProps {
+  cards: ReadonlyArray<PressCardItemProps>
```
palmenhq: 😍

PR #264, opened 2019-07-24
Blobs

axelbacklund:
Adding blob option to spacing elements!


src/utils/BlobDirection.ts
```diff
@@ -0,0 +1,35 @@
+export type BlobDirection = 'none' | 'up' | 'down'
+export interface BaseBlobProps {
```
palmenhq: i loled, this name is amazing 😂

src/utils/BlobDirection.ts
```diff
@@ -0,0 +1,35 @@
...
+    top: 'auto',
+    bottom: 0,
+  },
+  down: {
+    top: 0,
+    bottom: 'auto',
+  },
+}
+
+const upBlobPathData: string =
```
palmenhq: not that it's bad to have it, but i think TS can infer the `string` type :)

axelbacklund: True!

axelbacklund:
Thanks!! 😃Just thinking, if we're removing the inherent spacing of blocks per Oscar's PR, maybe it will be enough for now using the spacer blocks?

axelbacklund:
Checked with the group, should be fine with just spacer blocks 👍

PR #268, opened 2019-07-26
More alignment options

axelbacklund:
Adding some requested alignment options to the image text block - text alignment and option to put button below all other content on mobile.

### DIFF
--- a/src/blocks/ImageTextBlock.tsx
+++ b/src/blocks/ImageTextBlock.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import styled from 'react-emotion'
+import MediaQuery from 'react-responsive'
 import {
   ContentWrapper,
   SectionWrapper,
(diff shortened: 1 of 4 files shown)


src/blocks/ImageTextBlock.tsx
```diff
@@ -137,14 +155,17 @@ interface ImageTextBlockProps extends BaseBlockProps {
...
   button_weight?: ButtonWeight
+  button_position_mobile?: 'above' | 'below'
 }
 
 export const ImageTextBlock: React.FunctionComponent<ImageTextBlockProps> = ({
   title_size,
   title,
   paragraph,
   text_position,
+  text_alignment_mobile,
```
palmenhq: is there a difference between `alignment` and `position`? Otherwise i think we should pick one of them

src/blocks/ImageTextBlock.tsx
```diff
@@ -158,16 +179,51 @@ export const ImageTextBlock: React.FunctionComponent<ImageTextBlockProps> = ({
   size,
   media_position,
   button_color,
+  button_size,
   button_weight,
+  button_position_mobile,
 }) => {
+  const getButton = () =>
```
palmenhq: i think we should move this out to be a "normal" react component :) so it's used like
```jsx
<AlignedButton branchLink={button_branch_link} size={button_size} ...blablabla />
```
 or something like that

src/blocks/ImageTextBlock.tsx
axelbacklund: Makes sense, changing to text_position_mobile

PR #289, opened 2019-08-06
Add script to avoid screen flicker in A/B testing

axelbacklund:


hedvigoscar:
I literally hate this so much. So, so much.

src/server/utils/optimize.ts
```diff
@@ -0,0 +1,6 @@
+export const optimizeSnippet = `<style>.async-hide { opacity: 0 !important} </style>
```
palmenhq: imo we should place this in the helmet `script`s instead

hedvigoscar: We made the decision to not integrate this script as we considered the effect it has to be worse than flickering. If we want to prevent the root issue we should integrate it server-side, probably.

palmenhq: Reasonable enough, can we close this then?

hedvigoscar: Yes.

PR #299, opened 2019-08-23
Partner & code cookies

palmenhq:
Allows us to set partner + code already on the web. Used in combination with HedvigInsurance/web-onboarding#142


hedvigoscar:
How does this play in relation to the already existing code to automatically consume discount codes in web-onboarding?

palmenhq:
@hedvigoscar it's really just an extension, so we can send them to hedvig.com/?code=blablabla and it'll stick in the web onboarding also. Web onboarding just sucks up the cookie if it exists and puts it in the state https://github.com/HedvigInsurance/web-onboarding/pull/142/files#diff-280df203a4c1546ca4c782ec73e286c4R113

PR #308, opened 2019-09-17
Fix missing metadata

gustafguner:
- Added the mandatory `og:url` meta tag
- Provided some mandatory twitter tags (used the OG content)
- Removed an unwanted extra `/` from both `canonical` and `og:url`


hedvigoscar:
Are these represented as Trello tickets? 🙂

PR #339, opened 2019-10-28
Extend video support

gustafguner:
Extending video support by adding support for video background and "video image" in the "Image Text Block". The loading of each video is deferred until it enters the viewport, and it is dispatched when it leaves the viewport.

### DIFF
--- a/src/blocks/ImageTextBlock/index.tsx
+++ b/src/blocks/ImageTextBlock/index.tsx
@@ -169,22 +169,6 @@ const ImageVideo = styled(DeferredVideo)({
   borderRadius: 0.01,
 })
 
-const Video = styled('video')({
-  width: '100%',
-  objectFit: 'cover',
-  transition: 'height 1500ms',
-  overflow: 'hidden',
-  borderRadius: 0.01,
-  position: 'absolute',
-  bottom: 0,
-  right: 0,
-  left: 0,
-  height: '100%',
-  '@media(min-width: 1500px)': {
-    height: '110vh',
-  },
-})
-
 interface State {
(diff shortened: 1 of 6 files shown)


src/blocks/ImageTextBlock/index.tsx
```diff
@@ -142,6 +170,14 @@ const ImageLink = styled('a')(
   }),
 )
 
+const ImageVideo = styled(DeferredVideo)({
+  width: '100%',
+  objectFit: 'cover',
```
palmenhq: This guy is a bit dangerous because the browser support is fairly poor still. Just double checking it looks *decent* also without this property. see https://caniuse.com/#search=object-fit

src/components/DeferredVideo.tsx
```diff
@@ -0,0 +1,100 @@
+import * as React from 'react'
+import styled from 'react-emotion'
+import VisibilitySensor from 'react-visibility-sensor'
+
+const Video = styled('video')({
+  width: '100%',
+  objectFit: 'cover',
```
palmenhq: same, just double checking :)

src/blocks/ImageTextBlock/index.tsx
gustafguner: Yes, I'm aware of that! I figured since it's already being used (such as in `HomeHeroBlock`) I could just continue with that. And I guess if it's looking decent even without it, it's not a problem that it's not working in some old browsers? 😄

palmenhq: Tots :)

PR #354, opened 2019-12-06
dependency lovin ❤️

palmenhq:
Here's some upgrades of dependencies, including emotion 10. Sorry for the massive PR, but i really didn't change much except the compatibility issues 🙈


src/blocks/ImageTextBlock/index.tsx
```diff
@@ -28,75 +28,62 @@ import { BackgroundVideo } from './BackgroundVideo'
...
-        : 'row',
-    justifyContent: 'space-between',
-    alignItems: 'center',
-    [TABLET_BP_DOWN]: {
-      flexDirection: 'column',
-    },
-  }),
-)
+const AlignableContentWrapper = styled(ContentWrapper)<{
+  textPosition: string
```
hedvigoscar: Should this use the `TextPosition`-type? 🙂

src/blocks/ImageTextBlock/index.tsx
```diff
@@ -107,68 +94,56 @@ const Paragraph = styled('div')(
...
-  }: {
-    alignment: string
-    displayOrder: 'top' | 'bottom'
-    hasLink?: boolean
-  }) => ({
-    width: hasLink ? '100%' : '40%',
-    flexShrink: hasLink ? 1 : 0,
+const ImageVideoWrapper = styled('div')<{
+  alignment: string
+  displayorder: 'top' | 'bottom'
```
hedvigoscar: This also changed case 🤔

src/components/GlobalStyles.tsx
```diff
@@ -1,82 +1,83 @@
+import { css } from '@emotion/core'
 import { colors, fonts, getCdnFontFaces } from '@hedviginsurance/brand'
-import * as React from 'react'
-import { injectGlobal } from 'react-emotion'
 
-export const GlobalStyles: React.FunctionComponent = () => {
-  // tslint:disable-next-line no-unused-expression
-  injectGlobal`
-${getCdnFontFaces()}
+export const globalStyles = () => css`
```
hedvigoscar: Does this need to be a function? 🙂

src/blocks/AnimatedPhoneBlock.tsx
```diff
@@ -50,14 +50,14 @@ const TextWrapper = styled('div')(
...
-  ({ displayOrder }: { displayOrder: 'top' | 'bottom' }) => ({
+const Title = styled('h2')<{ displayorder: 'top' | 'bottom' }>(
+  ({ displayorder }) => ({
     fontSize: '4.5rem',
     wordWrap: 'break-word',
     width: '100%',
     [TABLET_BP_DOWN]: {
       fontSize: '3.75rem',
-      marginTop: displayOrder === 'top' ? '3rem' : '1.414rem',
+      marginTop: displayorder === 'top' ? '3rem' : '1.414rem',
```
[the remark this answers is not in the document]
palmenhq: yes, new react complains about the property `displayOrder` for some reason :)

src/blocks/ImageTextBlock/index.tsx
palmenhq: same about this one, react complains about the camel case and suggests this name 🤷‍♀️

src/components/GlobalStyles.tsx
palmenhq: no, but i like the immutability

.babelrc.js
```diff
@@ -8,6 +8,8 @@ module.exports = {
     'react-hot-loader/babel',
     '@babel/plugin-syntax-dynamic-import',
     '@babel/plugin-proposal-class-properties',
+    '@babel/plugin-proposal-optional-chaining',
+    '@babel/plugin-proposal-nullish-coalescing-operator',
```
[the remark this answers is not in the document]
palmenhq: Yes, it does since 1.19 which i've upgraded it to :)

src/components/GlobalStyles.tsx
hedvigoscar: Will it not be evaluated once at module evaluation time and then be constant?

src/blocks/AnimatedPhoneBlock.tsx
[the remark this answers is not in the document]
hedvigoscar: Fair enough, we can disregard all of these 🙂

src/components/GlobalStyles.tsx
palmenhq: yes, but you can still modify it after. I guess its a matter of preference, i like the thought of not being able to modify stuff. But i can make it mutable

hedvigoscar: It's not super-important to me. I think as a rule if someone starts mutating the output of a module we have to bust them in PRs anyways as that's highly illegal 👮

palmenhq: haha yes that's fair

PR #361, opened 2019-12-13
Fix overriden CTA links in header being missing on mobile

hedvigoscar:


src/blocks/HeaderBlock/index.tsx
```diff
@@ -220,6 +220,26 @@ class Header extends React.PureComponent<
                   {!isOpen && (
                     <>
                       {(() => {
+                        if (
+                          this.props.override_cta_link &&
+                          this.props.override_cta_link.cached_url
```
palmenhq: what about `override_mobile_header_cta_label`?

PR #372, opened 2020-02-16
Add ToC redirects

palmenhq:
So we get prettier urls and don't have to repoint links every time we change our terms

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "author": "Johan Palmfjord <johan.palmfjord@hedvig.com>",
   "types": "./types.d.ts",
   "engines": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -7,8 +7,8 @@
   "author": "Johan Palmfjord <johan.palmfjord@hedvig.com>",
   "types": "./types.d.ts",
   "engines": {
-    "node": "10.14.1",
-    "yarn": "1.19.2"
+    "node": "10.19.0",
+    "yarn": "1.22.0"
```
hedvigoscar: ❤️

PR #374, opened 2020-03-03
Spring cleaning

palmenhq:


gustaveen:
Good job! 👏  Tried to get the new edit view working locally, do you need to enable it somehow?

palmenhq:
@gustaveen thanks! It _should_ work out of the box 🤔 it only works inside the storyblok editor though, but i'm assuming that's where you tried?

gustaveen:
Never mind, I looked at the wrong place 😄 It works perfectly fine 🙌 @palmenhq

PR #375, opened 2020-03-03
Reskinning color prep + column & headline block

palmenhq:
<strike>we need to merge #374 before this, but then lets do it!</strike>


src/blocks/TitleCtaBlock.tsx
```diff
@@ -108,7 +108,7 @@ export const TitleCtaBlock: React.FunctionComponent<TitleCtaBlockInterface> = ({
                 styleType={button_type}
                 size="sm"
                 weight={button_weight}
-                color={button_color && button_color.color}
+                color={button_color?.color}
```
gustaveen: Ah, nice with optional chaining 🙏

PR #378, opened 2020-03-06
Brand pivot FooterBlock

gustaveen:
`SectionWrapper` now has a `brandPivot` prop that changes the font-family:

Also bumped the brand package version that includes the update colors.

Screenshot 2020-03-06 at 10 56 24

### DIFF
--- a/src/blocks/FooterBlockBrandPivot.tsx
+++ b/src/blocks/FooterBlockBrandPivot.tsx
@@ -1,0 +1,246 @@
+import styled from '@emotion/styled'
+import * as React from 'react'
+import {
+  ContentWrapper,
+  SectionWrapper,
+  TABLET_BP_DOWN,
+  TABLET_BP_UP,
+  SITE_MAX_WIDTH,
+  CONTENT_GUTTER,
+  CONTENT_GUTTER_MOBILE,
+  MOBILE_BP_UP,
+} from '../components/blockHelpers'
+import { ContextContainer } from '../components/containers/ContextContainer'
+import { GlobalStoryContainer } from '../storyblok/StoryContainer'
+import { getStoryblokLinkUrl } from '../utils/storyblok'
+import { BaseBlockProps } from './BaseBlockProps'
+import { colorsV3 } from '@hedviginsurance/brand'
+import { HedvigH } from '../components/icons/HedvigH'
+
+const BP_DOWN = '@media (max-width: 600px)'
+
+const FooterWrapper = styled('div')({
+  padding: `5rem 0`,
+
+  [MOBILE_BP_UP]: {
+    padding: `4.25rem 0 12rem`,
+  },
+})
+
+const IconWrapper = styled('div')({
+  display: 'flex',
+  justifyContent: 'flex-end',
+  margin: '0 auto 3.125rem',
+  paddingLeft: CONTENT_GUTTER_MOBILE,
+  paddingRight: CONTENT_GUTTER_MOBILE,
+
+  [MOBILE_BP_UP]: {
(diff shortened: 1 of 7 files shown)


gustaveen:
Ah I might have forgotten something, do you need to do a sync with storyblok to update the `components.json` as well? @palmenhq

src/blocks/FooterBlockBrandPivot.tsx
```diff
@@ -0,0 +1,246 @@
...
+                <FooterInnerWrapper>
+                  <LinkTextWrapper>
+                    {globalStory.content.get_started &&
+                    globalStory.content.get_started.length &&
+                    globalStory.content.company &&
+                    globalStory.content.company.length &&
+                    globalStory.content.legal &&
+                    globalStory.content.legal.length &&
+                    globalStory.content.social &&
+                    globalStory.content.social.length ? (
```
palmenhq: maybe we could make this slightly prettier with optional chaining?

something like

```js
globalStory.content.get_started?.length
&& globalStory.content.company?.length 
```

src/blocks/FooterBlockBrandPivot.tsx
```diff
@@ -0,0 +1,246 @@
...
+                      <LinksColumnsWrapper>
+                        <LinkColumn>
+                          <ColumnHeader>
+                            {context.lang === 'sv'
+                              ? 'Hemförsäkring'
+                              : 'Home insurance'}
```
palmenhq: I realize this wont go well with `no` :(

src/blocks/FooterBlockBrandPivot.tsx
```diff
@@ -0,0 +1,246 @@
...
+                        </LinkColumn>
+                      </LinksColumnsWrapper>
+                    )}
+
+                    <FooterFooter
+                      dangerouslySetInnerHTML={{
+                        __html:
+                          globalStory.content.footer_paragraph &&
+                          globalStory.content.footer_paragraph.html,
+                      }}
```
palmenhq: same, i think this could just be optional chaining now :)

src/blocks/FooterBlockBrandPivot.tsx
```diff
@@ -0,0 +1,246 @@
...
+      paddingBottom: '2rem',
+    },
+  },
+})
+const Link = styled('a')({
+  display: 'block',
+  color: 'inherit',
+  fontSize: '0.875rem',
+  textDecoration: 'none',
+  marginBottom: '0.8625rem',
```
palmenhq: for these really odd numbers, what's your thoughts, could it be better to just use `px`es?

palmenhq (commented):
looks amaze!
But yes, you will need to download your footer block to `components.json` :)

src/blocks/FooterBlockBrandPivot.tsx
gustaveen: Indeed!

src/blocks/FooterBlockBrandPivot.tsx
gustaveen: I hear you. I don't have a very strong opinon on that, I just tend to calculate size from design like 14/16 to get rem sizes.

palmenhq (dismissed):
Wooh!

src/blocks/FooterBlockBrandPivot.tsx
gustaveen: Very true, l'll have a look at that in my norway setup PR

gustaveen:
Updated the PR now :) @palmenhq

palmenhq (approved):
♻️

PR #380, opened 2020-03-09
Bullet point block

gustaveen:
I've tried to make this block more flexible so you can either have an image or an icon for each item. 
For icons you need to check a checkbox to get the correct layout. 

Screenshot 2020-03-09 at 15 49 04

Screenshot 2020-03-09 at 15 49 37

### DIFF
--- a/src/blocks/BulletPointBlockBrandPivot.tsx
+++ b/src/blocks/BulletPointBlockBrandPivot.tsx
@@ -1,0 +1,161 @@
+import styled from '@emotion/styled'
+import * as React from 'react'
+import {
+  ContentWrapper,
+  MOBILE_BP_DOWN,
+  MOBILE_BP_UP,
+  TABLET_BP_UP,
+  LAPTOP_BP_UP,
+  SectionWrapper,
+} from '../components/blockHelpers'
+import { DeferredImage } from '../components/DeferredImage'
+import { getStoryblokImage, Image } from '../utils/storyblok'
+import { BaseBlockProps, MarkdownHtmlComponent } from './BaseBlockProps'
+import { colorsV3 } from '@hedviginsurance/brand'
+
+const BulletPointSectionWrapper = styled(SectionWrapper)({
+  overflowX: 'hidden',
+})
+
+const InnerWrapper = styled('div')({
+  display: 'flex',
+  flexWrap: 'wrap',
+  minWidth: '100%',
+
+  [MOBILE_BP_DOWN]: {
+    paddingLeft: '8px',
+    paddingRight: '8px',
+  },
+
+  [MOBILE_BP_UP]: {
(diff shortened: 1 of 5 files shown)


src/blocks/BulletPointBlockBrandPivot.tsx
```diff
@@ -0,0 +1,167 @@
...
+    maxWidth: 'none',
+    maxHeight: 'none',
+    width: 'auto',
+    marginRight: 0,
+    marginBottom: '24px',
+  },
+}))
+
+const BulletPointImage = styled(DeferredImage)<{
+  iconlayout: boolean
```
palmenhq: what about camel casing `iconLayout`? :)

src/blocks/BulletPointBlockBrandPivot.tsx
```diff
@@ -0,0 +1,167 @@
...
+    paddingLeft: '8px',
+    paddingRight: '8px',
+  },
+
+  [MOBILE_BP_UP]: {
+    marginLeft: '-40px',
+  },
+
+  [LAPTOP_BP_UP]: {
+    marginLeft: '-80px',
```
palmenhq: Is there a reason for all the `px` units here?

src/blocks/BulletPointBlockBrandPivot.tsx
gustaveen: Yes please, I'll fix :)

src/blocks/BulletPointBlockBrandPivot.tsx
gustaveen: Ah sorry, I should have written a comment about that :) When working with column layouts I like using this margin-left-hack, where you offset the container equally as the margin-left on the items them self. And I also noticed that we couldn't have too much spacing in tablet sizes.

Or did you mean that we should use rems units? @palmenhq

gustaveen: I think I'll have a look at building a grid component later on, using css grid with flexbox fallback 😊

palmenhq: Yes i was thinking more about whether we should use `rem`s or not really :)

That sounds amazing! Maybe something we could reuse in more places? such as the web onboarding or hedgehog

palmenhq (approved):
Amaze! Just questioning the px, but i don't think its that critical.

src/blocks/BulletPointBlockBrandPivot.tsx
gustaveen: I hear you, I read up a bit on the rem vs px battle today. For font-sizes it's still a no-brainer. But I think we can have look into if we want to use it for all margins as well 👍

PR #381, opened 2020-03-12
Add storybook

gustaveen:
## What
Adding storybook setup

## Why
To enable smoother local development workflow and in the end an ui library of all our components


package.json
```diff
@@ -20,6 +20,7 @@
     "@segment/snippet": "^4.4.0",
     "@segment/utm-params": "^2.0.0",
     "@sentry/node": "^4.3.0",
+    "@types/storybook-react-router": "^1.0.1",
```
palmenhq: Imo `@types` should go in `devDependencies` :)

palmenhq (commented):
Boom! Amaze!

package.json
gustaveen: Ah of course, my bad!

palmenhq (approved):
✅✅✅

PR #382, opened 2020-03-12
Add Banner block

gustaveen:
## Why
We need a way to direct people to our blog post about Corona on the top of the startpage

## What
A new block with a markdown editor

Screenshot 2020-03-12 at 14 04 42


palmenhq (approved):
🦠

PR #383, opened 2020-03-12
Add brand pivot max-width

gustaveen:


src/components/blockHelpers.tsx
```diff
@@ -37,6 +37,11 @@ export const CONTENT_MAX_WIDTH = {
   },
 }
 
+// New content max width
```
palmenhq: do we need this comment? 😬

src/components/blockHelpers.tsx
```diff
@@ -365,7 +373,11 @@ export const ContentWrapper: React.FC<ContentWrapperProps> = ({
     }}
   >
     {({ isVisible }) => (
-      <ContentWrapperStyled visible={index <= 1 || isVisible} {...props}>
+      <ContentWrapperStyled
+        visible={index <= 1 || isVisible}
+        brandPivot
```
palmenhq: `brandPivot={brandPivot}` ?

palmenhq (commented):
I feel like there's more blocks to change the `ContentWrapper` for but maybe not?

src/components/blockHelpers.tsx
gustaveen: Do we have a prefer to write the longer boolean syntax? :)

src/components/blockHelpers.tsx
gustaveen: Hehe nope, not necessary :) Just a reminder for us to remove the old max width later on.

src/components/blockHelpers.tsx
gustaveen: Yep, of course you need to do that when you pass down a prop, what was I thinking 😆

gustaveen:
Did as we discussed @palmenhq and inverted the variable names 👍

palmenhq (approved):
wooh!

PR #387, opened 2020-03-16
Feature/brand pivot hero block

gustaveen:
## New hero block

hero


src/blocks/HeroImageBlockBrandPivot/HeroImageBlockBrandPivot.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+  image: '',
+  image_mobile: '',
+}
+
+const image = 'https://cdn.hedvig.com/www/referrals/referrals-clean.png'
+
+export const Default = () => (
+  <HeroImageBlockBrandPivot
+    {...heroProps}
+    color={select('color', colors, colors.standardInverse)}
```
gustaveen: Having a hard time figuring out how to pass in color options to `addon-knobs` without getting type errors. It basically needs a string and not and object. Do you have any suggestion @palmenhq

src/blocks/HeroImageBlockBrandPivot/HeroImageBlockBrandPivot.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+  gray700: {
+    _uid: '6ecde11d-ba0a-48fb-9b7b-e6dbf31415d9',
+    color: 'gray700',
+    plugin: 'hedvig_minimal_color_picker',
+  },
+  gray500Inverse: {
+    _uid: '6ecde11d-ba0a-48fb-9b7b-e6dbf31415d9',
+    color: 'gray500-inverse',
+    plugin: 'hedvig_minimal_color_picker',
+  },
```
palmenhq: these'll be used in more places no? should we extract it to somewhere where its reusable?

storyblok/components.json
```diff
@@ -55,7 +55,9 @@
           "type": "bloks",
```
palmenhq: we should run prettier on this file :)

palmenhq (commented):
wooh!

src/blocks/HeroImageBlockBrandPivot/HeroImageBlockBrandPivot.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+  image: '',
+  image_mobile: '',
+}
+
+const image = 'https://cdn.hedvig.com/www/referrals/referrals-clean.png'
+
+export const Default = () => (
+  <HeroImageBlockBrandPivot
+    {...heroProps}
+    color={select('color', colors as any, colors.standardInverse as any)}
```
palmenhq: maybe we could instead do something like
```ts
color={
colors[select('color', Object.keys(colors), 'standard-inverse')]
}
```

gustaveen: Yes that's what I wanted too 🙂 But it needs a the `MinimalColorComponent` object as an option

storyblok/components.json
gustaveen: I saw that i messed up with this file and commited removed blocks, will fix 🙈

src/blocks/HeroImageBlockBrandPivot/HeroImageBlockBrandPivot.stories.tsx
gustaveen: that might be good! what do you think about some utils for storybook?

src/blocks/HeroImageBlockBrandPivot/HeroImageBlockBrandPivot.stories.tsx
palmenhq: Yes, isn't that what this'd do? Or not really?

src/blocks/HeroImageBlockBrandPivot/HeroImageBlockBrandPivot.stories.tsx
gustaveen: I've created a `utils/storybook.ts` and moved it there.

gustaveen:
Updated now @palmenhq Thanks a lot for the typescript support ❤️

src/blocks/HeroImageBlockBrandPivot/HeroImageBlockBrandPivot.tsx
```diff
@@ -0,0 +1,178 @@
...
+  headline,
+  text,
+  text_color,
+  image,
+  index,
+  image_mobile,
+  use_display_font = false,
+  show_hedvig_wordmark,
+}) => {
+  const isTablet = useMediaQuery({ maxWidth: 801 })
```
palmenhq: Does this work though? Considering we're using SSR. I think it'd be better to use client-safe css media queries

palmenhq (commented):
Amaze! Just one thought

src/blocks/HeroImageBlockBrandPivot/HeroImageBlockBrandPivot.tsx
gustaveen: Ah, good catch, that might be a problem. Not sure how this works under the hood but if would it work if you use the component syntax like we do in other places?:
```suggestion
<MediaQuery query="(max-width: 801px)"> 
   {...}
</MediaQuery>
```

gustaveen: Let's just stick to good ol' css https://github.com/HedvigInsurance/web-next/pull/387/commits/0c8fbe335f42e702c832af5e93804aa5008d8834 😄

palmenhq (approved):
amaze!

PR #388, opened 2020-03-17
Feature/header brand pivot

gustaveen:
## New header block

Refactored Header to a functional component so we can use internal state and avoid using refs for updating styles when scrolling. 

header

### Not in this PR
Haven't adjusted the mobile menu since designs are not a 100% completed


src/blocks/HeaderBlockBrandPivot/index.tsx
```diff
@@ -0,0 +1,366 @@
...
+    transition: `right ${TOGGLE_TRANSITION_TIME}ms`,
+    color: colorsV3.gray900,
+    overflow: 'scroll',
+    WebkitOverflowScrolling: 'touch',
+  },
+}))
+
+const LogoLink = styled('a')({
+  display: 'inline-flex',
+  paddingTop: 3, // fix to push down logo a little or it looks unbalanced
```
palmenhq: do we need this with the new logo?

src/blocks/HeaderBlockBrandPivot/index.tsx
```diff
@@ -0,0 +1,366 @@
...
+
+enum InverseColors {
+  DEFAULT = 'standard-inverse',
+  INVERSE = 'standard',
+}
+
+export const Header: React.FC<{ story: GlobalStory } & HeaderBlockProps> = (
+  props,
+) => {
+  const [isBelowThreshold, setScrollPosition] = React.useState<boolean>(
```
palmenhq: I think the setter here should match the state, so `setIsBelowThreshold`

src/blocks/HeaderBlockBrandPivot/index.tsx
```diff
@@ -0,0 +1,366 @@
...
+    minimalColorComponentColors | undefined
+  >(props.cta_color?.color)
+
+  React.useEffect(() => {
+    if (!props.is_transparent) {
+      return
+    }
+
+    setTimeout(() => {
+      updateHeader() // update the scroll animation after mount because the server doesn't know scroll position
```
palmenhq: isn't `useEffect` triggered after mount already? if it isn't then `setTimeout` is a bit of a tankevurpa anyways i believe because `setTimeout` works in node also. If we want something on the client side only we should probably check for `typeof window !== 'undefined'` or something simliar, but considering we're using `window.addEventListener` below i'm quite sure this does not run on the server. Maybe we should run this on `window.onload` or something rather? or just not use `setTimeout` in the first place. Or at least remove this comment since it's not true ;)

src/components/icons/HedvigLogotype.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const Svg = styled('svg')({
+  fill: 'currentColor',
+})
+
+export const HedvigLogotype: React.FunctionComponent<{ width: number }> = ({
+  width,
+}) => (
+  <Svg width={String(width)} viewBox="0 0 439 124">
+    <g fill-rule="nonzero">
```
palmenhq: this should probably be `fillRule`

palmenhq (commented):
love it 😍

src/blocks/HeaderBlockBrandPivot/index.tsx
gustaveen: Ah good catch! 🙌 It actually skews that logo since the height is set to auto so let's remove it. Will  adjust it later if we see that needs to be re-aligned in mobile views

src/blocks/HeaderBlockBrandPivot/index.tsx
gustaveen: Word!

src/blocks/HeaderBlockBrandPivot/index.tsx
gustaveen: Yes you're absolutely right, the timeout should be removed as well. Made sure we set `isBelowThreshold` state to false so we do the check in the hook instead and that fixed the issue when you're reload and are already scrolled down on the page.

palmenhq (approved):
noice!

PR #389, opened 2020-03-19
Cleaner imports 

gustaveen:
## What 
Add `esModuleInterop` rule to tsconfig for cleaner imports.
so instead of having to write `import * as React from 'react'` you can just write `import React from 'react`

## Why
Less code, easier to read and you can specify imports like `import React, { useEffect }  from 'react`

For more background this article highlights the problem: https://itnext.io/great-import-schism-typescript-confusion-around-imports-explained-d512fc6769c2


palmenhq (dismissed):
Nice, I always thought this was "bad practice" but apparently not! Just an idea.. should we fix all `import * as X`?

palmenhq (approved):
🙌

PR #391, opened 2020-03-23
Add hreflang fields to story page

gustaveen:
Now we get alternates from storyblok instead.

### DIFF
--- a/storyblok/components.json
+++ b/storyblok/components.json
@@ -3501,6 +3501,22 @@
           "type": "image",
           "pos": 9,
           "display_name": "SEO Meta og:image"
+        },
+        "hreflang_en": {
+          "type": "multilink",
+          "link_scope": "en/"
+        },
+        "hreflang_sv": {
+          "type": "multilink",
+          "link_scope": "sv/"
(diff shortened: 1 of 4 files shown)


palmenhq (dismissed):
Awesome!

storyblok/components.json
```diff
@@ -3501,6 +3501,22 @@
...
+          "type": "multilink",
+          "link_scope": "sv/"
+        },
+        "hreflang_no": {
+          "type": "multilink",
+          "link_scope": "no/"
+        },
+        "Hreflang": {
+          "type": "section",
+          "keys": ["hreflang_en", "hreflang_sv", "hreflang_no"]
```
palmenhq, line 3519: Can we kill of this stuff now maybe? Seems it is not used in the `<head>` anymore? (and all the rests from it)

palmenhq (dismissed):
awesome!

storyblok/components.json
gustaveen: Oops, of course! Thanks for catching that!

PR #393, opened 2020-03-23
New quote block

gustaveen:
New quote block have support for multiple quotes so we can add a slider or fade animation later on.

quote


src/blocks/QuoteBlockBrandPivot/QuoteBlockBrandPivot.tsx
```diff
@@ -0,0 +1,77 @@
...
+  quotes,
+  index,
+}) => (
+  <SectionWrapper brandPivot colorComponent={color}>
+    <ContentWrapper brandPivot index={index}>
+      {quotes.map((quote) => (
+        <QuoteWrapper largeQuote={quote.large_quote}>
+          <Quote largeQuote={quote.large_quote}>{quote.quote}</Quote>
+          <Cite colorComponent={color}>
+            {'– '}
```
palmenhq: Do we need to add this? I'm thinking its easy to put into the cms, and that means the content editors can remove this if they don't want it for some reason.

gustaveen: True, let's remove that 👍

gustaveen:
Update now @palmenhq :)

PR #397, opened 2020-03-25
Wordmark for title in image text block

axelbacklund:


src/blocks/ImageTextBlockBrandPivot/index.tsx
```diff
@@ -178,11 +179,31 @@ const ImageVideo = styled(DeferredVideo)({
...
+    width: '2rem',
+    height: '2rem',
+  },
+
+  [TABLET_BP_DOWN]: {
+    ['svg']: {
+      width: '1.25rem',
+      height: '1.25rem',
+    },
+  },
```
gustaveen: When refactoring I've seen we have a mix of max-width and min-width media queries. I think it's nice if we use a mobile first approach and use min-width media queries :)
```suggestion
  ['svg']: {
    width: '1.25rem',
    height: '1.25rem',
  },

  [TABLET_BP_UP]: {
    ['svg']: {
      width: '2rem',
      height: '2rem',
    },
  },
```

PR #401, opened 2020-03-31
Perils block

gustaveen:
## Perils for market-web! 

Didn't want to make this PR larger even though there's some more stuff to fix

perils

### DIFF
--- a/src/components/ButtonBrandPivot/Button.tsx
+++ b/src/components/ButtonBrandPivot/Button.tsx
@@ -75,7 +75,7 @@ export const ButtonBrandPivot = styled('button')<
       padding: buttonSizes[size],
       borderRadius: 8,
       border: `1px solid ${colorStyles.background}`,
-      fontFamily: fonts.FAVORIT,
+      fontFamily: 'inherit',
       textDecoration: 'none',
       textAlign: 'center',
       fontWeight: 'normal',
--- a/src/components/ButtonBrandPivot/Button.tsx
+++ b/src/components/ButtonBrandPivot/Button.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled'
-import { colorsV3, fonts } from '@hedviginsurance/brand'
+import { colorsV3 } from '@hedviginsurance/brand'
 import {
   colorComponentColors,
   minimalColorComponentColors,
--- a/src/components/GlobalStyles.tsx
+++ b/src/components/GlobalStyles.tsx
@@ -1,7 +1,12 @@
 import { css } from '@emotion/core'
-import { colors, fonts, getCdnFontFaces } from '@hedviginsurance/brand'
+import {
+  colors,
+  colorsV3,
+  fonts,
+  getCdnFontFaces,
+} from '@hedviginsurance/brand'
 
-export const globalStyles = () => css`
+export const globalStyles = (brandPivot: boolean) => css`
   ${getCdnFontFaces()}
 
   * {
@@ -9,12 +14,12 @@ export const globalStyles = () => css`
   }
 
   body {
-    font-family: ${fonts.CIRCULAR}, sans-serif;
+    font-family: ${brandPivot ? fonts.FAVORIT : fonts.CIRCULAR}, sans-serif;
     font-size: 16px;
     line-height: 1.5;
     margin: 0;
     padding: 0;
-    color: ${colors.OFF_BLACK};
+    color: ${brandPivot ? colorsV3.gray900 : colors.OFF_BLACK};
   }
 
   h1,
@@ -23,7 +28,7 @@ export const globalStyles = () => css`
   h4,
   h5,
   h6 {
-    font-family: ${fonts.GEOMANIST}, sans-serif;
+    font-family: ${brandPivot ? fonts.FAVORIT : fonts.GEOMANIST}, sans-serif;
     font-kerning: none;
     margin: 1.414rem 0 0.5rem;
     font-weight: inherit;
--- a/src/components/GlobalStyles.tsx
+++ b/src/components/GlobalStyles.tsx
@@ -6,7 +6,7 @@ import {
   getCdnFontFaces,
 } from '@hedviginsurance/brand'
 
-export const globalStyles = (brandPivot: boolean) => css`
+export const globalStyles = css`
   ${getCdnFontFaces()}
 
   * {
@@ -14,12 +14,12 @@ export const globalStyles = (brandPivot: boolean) => css`
   }
 
   body {
-    font-family: ${brandPivot ? fonts.FAVORIT : fonts.CIRCULAR}, sans-serif;
+    font-family: ${fonts.CIRCULAR}, sans-serif;
     font-size: 16px;
     line-height: 1.5;
     margin: 0;
     padding: 0;
-    color: ${brandPivot ? colorsV3.gray900 : colors.OFF_BLACK};
+    color: ${colors.OFF_BLACK};
   }
 
   h1,
@@ -28,7 +28,7 @@ export const globalStyles = (brandPivot: boolean) => css`
   h4,
   h5,
   h6 {
-    font-family: ${brandPivot ? fonts.FAVORIT : fonts.GEOMANIST}, sans-serif;
+    font-family: ${fonts.GEOMANIST}, sans-serif;
     font-kerning: none;
     margin: 1.414rem 0 0.5rem;
     font-weight: inherit;
@@ -86,5 +86,23 @@ export const globalStyles = (brandPivot: boolean) => css`
   input,
   button {
     font-size: inherit;
+  }
+`
+
+export const globalStylesBrandPivot = css`
+  ${globalStyles}
+
+  body {
+    font-family: ${fonts.FAVORIT}, sans-serif;
+    color: ${colorsV3.gray900};
+  }
+
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    font-family: ${fonts.FAVORIT}, sans-serif;
   }
 `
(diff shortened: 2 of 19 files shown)


src/components/GlobalStyles.tsx
```diff
@@ -1,20 +1,25 @@
...
-import { colors, fonts, getCdnFontFaces } from '@hedviginsurance/brand'
+import {
+  colors,
+  colorsV3,
+  fonts,
+  getCdnFontFaces,
+} from '@hedviginsurance/brand'
 
-export const globalStyles = () => css`
+export const globalStyles = (brandPivot: boolean) => css`
```
palmenhq: Rather than making a flag out of this, could we have separate `globalStylesBrandPivot`'? so we do an extension of the `globalStyles` for brand pivot. If possible i think we should try to avoid bool flags like this, looking at where the function is used its quite hard to make out what `true` or `false` means

src/components/Perils/PerilItem/PerilItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+  min-height: 3.75rem;
+  padding: 0.5rem 1rem 0.5rem 0.375rem;
+  color: inherit;
+  font-family: inherit;
+  border-radius: 0.375rem;
+  border: 0;
+  background-color: ${colorsV3.gray100};
+  cursor: pointer;
+  transition: all 150ms ease-in-out;
+  -webkit-appearance: none;
```
palmenhq: Is there any particular reason we're only targeting webkit? Or maybe we should just make this `appearance: none;`? I'm pretty sure emotion would take care of the prefixing if so

src/components/Perils/PerilItem/PerilItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  :focus {
+    outline: none;
+  }
+
+  &:focus {
+    box-shadow: none;
+    outline: none;
+  }
```
palmenhq: `:focus` or `&focus`?

src/components/Perils/PerilItem/PerilItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  ${TABLET_BP_UP} {
+    flex-direction: column;
+    align-items: flex-start;
+    padding: 0.5rem;
+
+    :hover {
```
palmenhq: same, shouldn't this really be `&:hover`? Otherwise if i'm not mistaken this might target childrens' hovers

src/components/Perils/PerilItem/PerilItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    fetchIcon()
+  }, [icon])
+
+  return (
+    <OuterContainer>
+      <Container>
+        {iconString && (
+          <IconWrapper dangerouslySetInnerHTML={{ __html: iconString }} />
```
palmenhq: since we know the `svg` is going to be 2rem x 2rem, maybe we should set that on the `IconWrapper` and always display it? Just for the full page not to flicker once the icons start popping in.

src/components/Perils/PerilItem/PerilItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+      }
+      const url = `https://graphql.dev.hedvigit.com${iconUrl}`
+      const iconResponse = await axios.get(url, {
+        withCredentials: false,
+      })
+      seticonString(iconResponse.data)
+    }
+
+    fetchIcon()
+  }, [icon])
```
palmenhq: should we maybe use `iconUrl` instead? Since that's ultimately what we care about. I think `useEffect` is not so good at deep-equaling objects, so we might risk checking on object identity rather than object structure, hence risk fetching this a few times

src/components/Perils/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  insuranceType: TypeOfContract
+}
+
+export const Perils: React.FC<Props> = ({ insuranceType }) => {
+  const [perils, setPerils] = useState<[] | Peril[]>([])
+
+  useEffect(() => {
+    const fetchPerils = async () => {
+      const url = `https://graphql.dev.hedvigit.com/graphql`
+      const data = `{"operationName":"Perils","variables":{"typeOfContract":"${insuranceType}","locale":"sv_SE"},"query":"query Perils($typeOfContract: TypeOfContract!, $locale: Locale!) {\\n perils(contractType: $typeOfContract, locale: $locale) {\\n title\\n description\\n covered\\n icon {\\n variants {\\n light {\\n svgUrl\\n }\\n }\\n }\\n }\\n}\\n"}`
```
palmenhq: could we maybe use `JSON.stringify` and do the actual query in a template string intsead? Imagine coming back to this in a few months and trying to make out what we're querying.

So this instead becomes:
```js
const data = JSON.stringify(
{
    "operationName": "Perils",
    "variables": {
        "typeOfContract": insuranceType,
        "locale": "sv_SE"
    },
    "query": `
        query Perils($typeOfContract: TypeOfContract!, $locale: Locale!) {
          perils(contractType: $typeOfContract, locale: $locale) {
            title
            description
            covered
            icon {
              variants {
                light {
                  svgUrl
                }
              }
            }
          }
        }
`
}
)
```

palmenhq: Acutally, i think you don't even need to stringify it with axios. it should be sufficient to pass the object directly, and axios will covert it to JSON automatically, if i remember correctly

src/components/blockHelpers.tsx
```diff
@@ -365,7 +366,11 @@ export const ContentWrapperStyled = styled('div')<{
     padding: '0 ' + CONTENT_GUTTER_MOBILE,
   },
 
-  ...(brandPivot ? CONTENT_MAX_WIDTH : CONTENT_MAX_WIDTH_DEPRECATED),
+  ...(brandPivot
+    ? fullWidth
+      ? SITE_MAX_WIDTH
+      : CONTENT_MAX_WIDTH
+    : CONTENT_MAX_WIDTH_DEPRECATED),
```
palmenhq: could we maybe move this to a function and name it? Ithink nested ternaries get really hard to read

src/components/Perils/PerilItem/PerilItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+}) => {
+  const [iconString, seticonString] = useState<string | null>(null)
+  const iconUrl: string = icon.variants.light.svgUrl
+
+  useEffect(() => {
+    const fetchIcon = async () => {
+      if (!iconUrl) {
+        return
+      }
+      const url = `https://graphql.dev.hedvigit.com${iconUrl}`
```
palmenhq: Shall we perhaps point this to production instead?

palmenhq (commented):
Just some thoughts, over all really good job!

src/components/Perils/PerilItem/PerilItem.tsx
gustaveen: Good thinking! 🙏

src/components/Perils/PerilItem/PerilItem.tsx
gustaveen: Yep hasty mistakes 🤦‍♂️

src/components/Perils/PerilItem/PerilItem.tsx
gustaveen: For sure! What is the prod endpoint though?

src/components/ButtonBrandPivot/Button.tsx
```diff
@@ -75,7 +75,7 @@ export const ButtonBrandPivot = styled('button')<
       padding: buttonSizes[size],
       borderRadius: 8,
       border: `1px solid ${colorStyles.background}`,
-      fontFamily: fonts.FAVORIT,
+      fontFamily: 'inherit',
```
palmenhq, line 78: clever!

src/components/Perils/PerilItem/PerilItem.tsx
palmenhq: https://giraffe.hedvig.com :)

src/components/Perils/data/usePerils.ts
```diff
@@ -0,0 +1,54 @@
+import axios from 'axios'
+import { useEffect, useState } from 'react'
+import { Peril, TypeOfContract } from '../types'
+
+export const usePerils = (insuranceType: TypeOfContract) => {
+  const [perils, setPerils] = useState<[] | Peril[]>([])
+
+  const fetchPerils = async () => {
+    const url = `https://graphql.dev.hedvigit.com/graphql`
```
palmenhq, line 9: I just noticed but i guess this should be the same as https://github.com/HedvigInsurance/web-next/pull/401/files#diff-01b4c9453b6be627da5698a4789ff9c3R131 - maybe we should put the host in a common place?

src/components/Perils/data/usePerils.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const data = {
+      operationName: 'Perils',
+      variables: {
+        typeOfContract: insuranceType,
+        locale: 'sv_SE',
```
palmenhq, line 15: One more thing i realised - we shouldn't forget to change this. We could probably use the [`Context`'s](https://github.com/HedvigInsurance/web-next/blob/master/src/components/containers/ContextContainer.tsx) `lang` property

palmenhq (dismissed):
Awesome! Just some ideas for the future, if we need to get this out quick

src/components/Perils/data/usePerils.ts
gustaveen: Word, I'll fix that in the next PR :)

palmenhq (approved):
awesome!

PR #402, opened 2020-04-01
Mobile header

axelbacklund:


src/server/utils/teamtailor.ts
```diff
@@ -92,7 +92,7 @@ export const initializeTeamtailorUsers = () =>
           e.status}", retrying after 5 sec...`,
       )
 
-      await new Promise((resolve) => setTimeout(resolve, 5000))
+      await new Promise((resolve) => setTimeout(resolve, 100))
```
gustaveen, line 95: Why do we need change this? :)

gustaveen:
Looks good! Just a few things that you might already work on:

- [ ] Make button full width in smaller viewports
- [ ] Cross icon should be white when menu is open
- [ ] The background of the header should be #121212 (the design is not correct)
- [ ] Make subitems smaller font size
- [ ] The H should be a wrapped in a link to /
- [ ] The fold out button is partly overlapped by the menu item, making in hard to click (see image below)

Screenshot 2020-04-02 at 10 50 33

src/server/utils/teamtailor.ts
axelbacklund: My mistake, shouldn't have been pushed 😅

src/blocks/HeaderBlockBrandPivot/index.tsx
```diff
@@ -215,69 +240,87 @@ export const Header: React.FC<{ story: GlobalStory } & HeaderBlockProps> = (
...
+                          href={
+                            '/' + (context.lang === 'sv' ? '' : context.lang)
+                          }
+                        >
+                          <HedvigLogotype width={98} />
+                        </LogoLink>
+                      )}
+                    </ContextContainer>
+                  </MediaQuery>
+                </RightContainer>
```
gustaveen, line 256: I was confused by this since it's to the left now haha, maybe change the name?
```suggestion
                </LeftContainer>
```

src/blocks/HeaderBlockBrandPivot/index.tsx
```diff
@@ -295,7 +264,8 @@ export const Header: React.FC<{ story: GlobalStory } & HeaderBlockProps> = (
                         <ButtonWrapper>
                           <ButtonLinkBrandPivot
                             styleType={props.cta_style}
-                            color={buttonColor}
+                            color={isOpen ? 'standard-inverse' : buttonColor}
```
gustaveen, line 267: I think we should still support the ability to change the button color

gustaveen:
I think we need to sync with Carl if we can remove the ability to change the mobile button independently.    And we also need to be able to have the button in mobile filled and in desktop outlined

src/blocks/HeaderBlockBrandPivot/index.tsx
```diff
@@ -207,41 +217,135 @@ export const Header: React.FC<{ story: GlobalStory } & HeaderBlockProps> = (
...
-                {!isOpen && (
-                  <>
+                <Menu open={isOpen}>
+                  {(props.story.content.header_menu_items ?? []).map(
+                    (menuItem) => (
+                      <MenuItem menuItem={menuItem} key={menuItem._uid} />
+                    ),
+                  )}
+
+                  <MediaQuery query="(max-width: 1000px)">
```
gustaveen, line 262: Ah, I realized now that it's the `<MediaQuery>` that causes the flickering. So if we want to show/hide depending on viewport we should wrap it in a div and just use css media querys to get rid of that problem :)

axelbacklund: Great 👍

gustaveen:
Awesome! Let's merge this bad boy!

PR #403, opened 2020-04-01
Perils block with compare feature and theme support

gustaveen:
Screenshot 2020-04-06 at 16 58 23


src/components/Perils/PerilItem/PerilItem.tsx
```diff
@@ -8,9 +8,11 @@ import {
 } from 'components/blockHelpers'
 import { PerilIcon } from 'components/Perils/types'
 import React, { useEffect, useState } from 'react'
+import { minimalColorComponentColors } from 'src/blocks/BaseBlockProps'
```
palmenhq, line 11: this import looks weird, does it work?

gustaveen: haha weird, yeah it does work somehow

PR #410, opened 2020-04-06
Add peril modal

gustaveen:
Refactored the modal from web-onboarding

modal

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "emotion": "^10.0.23",
     "emotion-server": "^10.0.17",
     "es6-promise": "^4.2.5",
+    "framer-motion": "^1.10.3",
     "ioredis": "^4.2.0",
     "is-ios": "^1.0.0",
     "js-cookie": "^2.2.0",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "emotion": "^10.0.23",
     "emotion-server": "^10.0.17",
     "es6-promise": "^4.2.5",
-    "framer-motion": "^1.10.3",
     "ioredis": "^4.2.0",
     "is-ios": "^1.0.0",
     "js-cookie": "^2.2.0",
--- a/src/components/Perils/PerilModal/PerilModal.tsx
+++ b/src/components/Perils/PerilModal/PerilModal.tsx
@@ -7,6 +7,7 @@ import { Cross } from 'components/icons/Cross'
 import { Minus } from 'components/icons/Minus'
 import { Modal, ModalProps } from 'components/Modal/Modal'
 import React from 'react'
+import { GlobalStory } from 'storyblok/StoryContainer'
 import { Peril } from '../types'
 
 const TRANSITION_MS = 250
@@ -15,6 +16,7 @@ interface PerilModalProps {
   perils: ReadonlyArray<Peril>
   currentPerilIndex: number
   setCurrentPeril: (perilIndex: number) => void
+  story?: GlobalStory
 }
 
 const ModalWrapper = styled('div')`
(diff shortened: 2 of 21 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
     "emotion": "^10.0.23",
     "emotion-server": "^10.0.17",
     "es6-promise": "^4.2.5",
+    "framer-motion": "^1.10.3",
```
palmenhq, line 38: Ouff, can we do this without framer? It is a HUGE library and i think it'd be good to omit it if we can. Alternatively, if we can split the code and load it asynchronously so that we don't include it in the initial bundle.

src/components/Perils/PerilModal/PerilModal.tsx
```diff
@@ -0,0 +1,269 @@
+import styled from '@emotion/styled'
+import { colorsV3, fonts } from '@hedviginsurance/brand'
+
+import { LAPTOP_BP_UP, TABLET_BP_UP } from 'components/blockHelpers'
+import { Chevron } from 'components/icons/Chevron'
+import { Cross } from 'components/icons/Cross'
+import { Minus } from 'components/icons/Minus'
+import { Modal, ModalProps } from 'components/Modal/Modal'
```
palmenhq, line 8: I think if we want to use `framer-motion` in the modal, we should code split that.

src/components/Perils/PerilModal/PerilModal.tsx
```diff
@@ -0,0 +1,269 @@
...
+          >
+            <Chevron direction="right" />
+          </DirectionButton>
+        </Header>
+        <Content>
+          <Column>
+            <Description>{currentPeril.description}</Description>
+            {currentPeril.info && (
+              <InfoBox>
+                <InfoBoxTitle>{'PLACEHOLDER'}</InfoBoxTitle>
```
palmenhq, line 235: 🤨

src/components/Perils/PerilModal/PerilModal.tsx
```diff
@@ -0,0 +1,269 @@
...
+                <InfoBoxBody>{currentPeril.info}</InfoBoxBody>
+              </InfoBox>
+            )}
+          </Column>
+          <Column>
+            <CoverageWrapper>
+              <CoverageList>
+                <CoverageListTitle>{'PLACEHOLDER'}</CoverageListTitle>
```
palmenhq, line 243: again?

src/components/Perils/PerilModal/PerilModal.tsx
```diff
@@ -0,0 +1,269 @@
...
+                  <CoverageListItem key={text}>
+                    <Cross size="0.75rem" />
+                    {text}
+                  </CoverageListItem>
+                ))}
+              </CoverageList>
+
+              {currentPeril.exceptions.length > 0 && (
+                <CoverageList>
+                  <CoverageListTitle>{'PLACEHOLDER'}</CoverageListTitle>
```
palmenhq, line 254: is it really supposed to say `placeholder`?

src/components/icons/Svg.tsx
```diff
@@ -1,14 +1,31 @@
...
+const getDirection = (direction: Direction) => {
+  return `
+  ${
+    direction === 'right'
+      ? 'rotate(-90deg)'
+      : direction === 'left'
+      ? 'rotate(90deg)'
+      : direction === 'up'
+      ? 'scaleY(-1)'
+      : direction === 'down' && 'scaleY(1)'
```
palmenhq, line 19: can we do this nicer? I think nested ternaries are a bit messy :/ Maybe using `matchly`?

package.json
gustaveen: Absolutely. I just took all the shortcuts I could with this but now that we have som extra time we can look at a more lightweight solution 🙌

src/components/Perils/PerilModal/PerilModal.tsx
gustaveen: Haha nope, I was just for me not to forget to add a global field for this in storyblok 😁

src/components/icons/Svg.tsx
gustaveen: I forgot your distaste for nested ternaries ;) `matchly` would be nice!

package.json
gustaveen: Removed it and tried to use react-spring. Updated that lib only to run in to some type and build errors. So resorted to plain css transitions for now, works fine I think 👍

PR #414, opened 2020-04-09
Fix/perils text size

gustaveen:


src/components/Perils/PerilModal/PerilModal.tsx
```diff
@@ -254,21 +254,6 @@ export const PerilModal: React.FC<PerilModalProps & ModalProps> = (props) => {
...
-                    {props.story?.content.peril_modal_exceptions_title ??
-                      'Exceptions'}
-                  </CoverageListTitle>
-                  {currentPeril.exceptions.map((text) => (
-                    <CoverageListItem key={text}>
-                      <Minus size="0.75rem" />
-                      {text}
-                    </CoverageListItem>
-                  ))}
-                </CoverageList>
```
palmenhq, line 270: we're skipping the exceptions? Why? Can we do that (as in like - is it legal?) ?

gustaveen: Carl got feedback from a few people that we shouldn't show it on the market web, since it might convert less and not other insurers do that. But in the web onboarding we should show it of course

palmenhq: Alrgihty, just double checking 👌

palmenhq (approved):

PR #422, opened 2020-04-15
Update/perils

gustaveen:
Screenshot 2020-04-15 at 13 39 33


src/components/Perils/data/usePerils.ts
```diff
@@ -7,7 +7,7 @@ export const usePerils = (insuranceType: TypeOfContract, locale: Locale) => {
   const [perils, setPerils] = useState<[] | Peril[]>([])
 
   const fetchPerils = async () => {
-    const url = `https://giraffe.hedvig.com/graphql`
+    const url = `https://graphql.dev.hedvigit.com/graphql`
```
palmenhq, line 10: hm. Should we maybe make this an environment variable instead?

gustaveen: yes, let's do it when I change it back tomorrow :)

PR #423, opened 2020-04-15
Feature/header market picker

gustaveen:
Will break out the Market Picker out to it's own component but this will do for now


src/blocks/HeaderBlockBrandPivot/HeaderTop.tsx
```diff
@@ -0,0 +1,279 @@
...
+  svg {
+    transform: rotate(45deg);
+  }
+`
+
+export const HeaderTop: React.FC<HeaderTopProps> = ({
+  transparent,
+  inverse,
+}) => {
+  const [openMarketPicker, toggleMarketPicker] = useState<boolean>(false)
```
palmenhq, line 217: this sounds like a function that'll open the market picker, maybe `isMarketPickerOpen` or `marketPickerOpen`?

PR #440, opened 2020-04-22
Fix cookie consent not unmounting

gustaveen:
This would not un mount when closing the cookie consent, only on reload when the cookie is already set.


palmenhq (approved):
hehe oops

PR #443, opened 2020-04-22
Put back segment snippet

palmenhq:


gustaveen (approved):
So segment handles some other tracking? Or why do we need to put it back?

palmenhq:
@gustaveen yes exactly, among other things. We're trying to replace it with GTM but apparently it wasn't this easy 🙈

PR #444, opened 2020-04-23
Refactor/global styling

gustaveen:
Adjust global styling to new theme.

Also added font-smoothing rule to body in order to make the font render thinner like the design. This is commonly used but not optimal since it has mixed support between browsers and might be discontinues in the future. The other option would be to try out using Favorit Light on the web but since we haven't bought that font and they were rather expensive we'll resort to this solution for now.


palmenhq (approved):
nice

PR #447, opened 2020-04-27
Fix/hreflang

gustaveen:
Address the following issues

- hreflang in `meta.tsx` for root gets a trailing slash and hence returns a 301 (/se/ -> /se)
- / redirects to /se/ and then to /se


palmenhq (approved):
nice!

PR #459, opened 2020-05-15
Peril adjustments

palmenhq:
Screenshot 2020-05-15 at 15 47 03
Screenshot 2020-05-19 at 09 15 19

### DIFF
--- a/src/blocks/PerilsBlock/PerilsBlock.tsx
+++ b/src/blocks/PerilsBlock/PerilsBlock.tsx
@@ -1,6 +1,9 @@
 import styled from '@emotion/styled'
 import { BrandPivotBaseBlockProps } from 'blocks/BaseBlockProps'
-import { ContentWrapper, SectionWrapper } from 'components/blockHelpers'
+import {
+  ContentWrapper as OriginalContentWrapper,
+  SectionWrapper,
+} from 'components/blockHelpers'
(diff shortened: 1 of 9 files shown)


gustaveen:
Good job! 🙏
Just saw that you forgot to remove `export const Description = styled.p` on line 99 `PerilItem.tsx` @palmenhq

gustaveen:
Also saw that I missed to add two bullets to the ticket. 

- The PerilItem should now be square

- The PerilsCollection has a narrower max-width (784px excluding padding)

But we can add a new ticket for this maybe

palmenhq:
@gustaveen haha nah i should've caught this in the design. Amended now anyways!

gustaveen:
Awesome! 👍Just saw that it has some offset of 2px and I can't see why 🤷‍♂️
Screenshot 2020-05-19 at 10 05 44

I also saw a really neat solution for making a div square without adding an extra container. If you add a pseudo element to `OuterContainer` like so:
```
&:before {
  content: "";
  padding-top: 100%;
}

But I'm 100% fine with this 🙌

gustaveen (dismissed):

palmenhq:
@gustaveen ehm wat. U sure? This is what i see. Which screen size are you looking at?

Screenshot 2020-05-19 at 10 34 51

gustaveen (approved):
Super duper! 🎉

PR #460, opened 2020-05-19
Introduce storyblok "hot" reloading instead of a hard page reload

palmenhq:
This'll cause the page content to update without reloading the page, both inline the storyblok editor and on a 5 sec interval if you add `?live_reload=true` to the url :)


gustaveen (approved):
Oh I like! Let's try it out 🙌

PR #466, opened 2020-05-25
Fix/image block

gustaveen:
Screenshot 2020-05-25 at 14 41 12

### DIFF
--- a/src/blocks/YoutubeVideoBlock/YoutubeVideoBlock.tsx
+++ b/src/blocks/YoutubeVideoBlock/YoutubeVideoBlock.tsx
@@ -1,18 +1,24 @@
 import { css, keyframes } from '@emotion/core'
 import styled from '@emotion/styled'
 import { colorsV3 } from '@hedviginsurance/brand'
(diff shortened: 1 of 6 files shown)


src/blocks/YoutubeVideoBlock/YoutubeVideoBlock.tsx
```diff
@@ -122,11 +127,13 @@ const VideoFrame = styled.iframe<{ width?: number; height?: number }>`
 interface YoutubeVideoBlockProps extends BrandPivotBaseBlockProps {
   overlay_image: Image
   video_id: string
+  caption?: MarkdownHtmlComponent
```
palmenhq, line 130: should we add this to the `storyblok/components.json`  file?

gustaveen: Yes! I didn't know what the best way was to get it locally but I will do a sync from staging as you suggested :)

gustaveen: The other way around haha, fetch to my local setup. Updated now anyway :)

PR #467, opened 2020-05-25
Remove branch link from all components

gustaveen:
- Make button link optional in ImageTextBlockBrandPivot. We're unable to sae and publish pages right now so this is urgent ⚠️
- Remove all branch links since we don't use them anymore


palmenhq (approved):
Should we also remove the `AppLink` component maybe?

PR #470, opened 2020-05-26
Add footer per default to all pages

gustaveen:
To omit the tedious work of changing and adding footer to all pages, let's add to all pages as default. Do not merge until all new footer content is done!

### DIFF
--- a/src/pages/StoryPage.tsx
+++ b/src/pages/StoryPage.tsx
@@ -1,9 +1,9 @@
-import { FooterBlock } from 'blocks/FooterBlock/FooterBlock'
 import SbEditable from 'patched/storyblok-react'
 import React from 'react'
 import Helmet from 'react-helmet-async'
 import { getBlockComponent } from '../blocks'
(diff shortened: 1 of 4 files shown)


src/pages/StoryPage.tsx
```diff
@@ -33,8 +34,21 @@ export const StoryPage: React.FunctionComponent<{ nonce?: string }> = ({
             )
           }
 
+          if (block.component === 'footer_block_brand_pivot') {
```
palmenhq, line 37: Instead of using this if i think we could just remove it from the map of block names/blocks right? I feel like that would be more clean.

gustaveen: Much better!

palmenhq (dismissed):
Awesome! Lets just fix that test :)

gustaveen:
Could you give me a hand with the test if you have time @palmenhq , can't really figure out how to fix it 🤔

palmenhq (approved):
awesome!

PR #505, opened 2020-07-29
Chore/remove unused bloks

gustaveen:
This is part of the Storyblok clean up i.e. removing unused and legacy components. When we did the redesign we copied the old component and made a new one named {Component}BrandPivot. So in most of the cases I will remove the none suffixed component and keep the BrandPivot instance.

I will try to make this in smaller batches to not have a massive PR in the end 😅


src/blocks/index.tsx
```diff
@@ -47,7 +46,7 @@ const blockComponents = {
   header_block: HeaderBlock,
   header_block_brand_pivot: HeaderBlockBrandPivot,
   about_us_hero_block: AboutUsHeroBlock,
-  accordion_block: AccordionBlock,
+  accordion_block: AccordionBlockBrandPivot,
```
gustaveen, line 47: Still have some page where this is used so will just map this to the new component until it's removed in the CMS

gustaveen:
> image
> 
> Love it 🧹

Haha yeah it feels soo nice! 😄

PR #519, opened 2020-08-25
Update readme instructions

gustaveen:


README.md
```diff
@@ -1,27 +1,92 @@
...
+#### Install Yarn
+
+1. Install `yarn` [https://classic.yarnpkg.com/en/docs/install/#mac-stable](https://classic.yarnpkg.com/en/docs/install/#mac-stable)
+
+#### Install yvm
+
+`yvm` is a neat yarn version manager just like `nvm` is for node.
+
+1. Install `yvm` [https://yvm.js.org/docs/overview#installation](https://yvm.js.org/docs/overview#installation)
+2. Run any yarn command and watch it automagically use the correct version of yarn. (Version specified in `package.json` or `.yvmrc`)
```
palmenhq, line 30: I'd like to challenge whether we should have this. Instead we should link to the [web contributing guidelines](https://github.com/HedvigInsurance/tech-stuff/blob/master/web/contributing.md) don't you think?

palmenhq (commented):
Really nice!

README.md
gustaveen: true, might be a bit overkill to repeat these steps in all repos. but we can add a link to the contributing guidelines maybe? @palmenhq

palmenhq: Yes this i think is a good idea!

palmenhq (dismissed):

palmenhq (approved):

PR #524, opened 2020-08-27
Fix/accordion block brand pivot

styrelius:
### This PR entails the following:
1) Replacing the name `AccordionBlockBandPivot` (yes the "r" was left out, it's not a typo) with `AccordionBlock` since the pivoting is done I guess?
2) Changing design of the `AccordionBlock` component slightly - remove border from `ExpanderWrapper` and add more padding to `AccordionContent` `div`. This was requested (from Ermir I think?) in the "FAQ" ticket in the project board for "Web iterations".

### Screenshots
**New design:**
image

**Old design:**
image

### DIFF
--- a/src/blocks/AccordionBlock/AccordionBlock.tsx
+++ b/src/blocks/AccordionBlock/AccordionBlock.tsx
@@ -100,8 +100,12 @@ const AccordionContent = styled('div')({
   overflowY: 'hidden',
 
   [TABLET_BP_UP]: {
-    padding: '3rem',
+    padding: '3rem 3rem 1.5rem 3rem',
     '& p': {
+      marginTop: 0,
+      marginBottom: '1rem',
+    },
+    '& p:last-child': {
       margin: 0,
     },
   },
--- a/src/blocks/AccordionBlock/AccordionBlock.tsx
+++ b/src/blocks/AccordionBlock/AccordionBlock.tsx
@@ -165,7 +165,7 @@ export interface AccordionBlockProps extends BrandPivotBaseBlockProps {
   accordions: ReadonlyArray<AccordionProps>
 }
 
-export const AccordionBlockBrandPivot: React.FunctionComponent<AccordionBlockProps> = ({
+export const AccordionBlock: React.FunctionComponent<AccordionBlockProps> = ({
   accordions,
   color,
   extra_styling,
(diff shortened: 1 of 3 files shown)


styrelius:
Please tell me if the description is too detailed :)

src/blocks/AccordionBlock/AccordionBlock.tsx
```diff
@@ -100,8 +100,10 @@ const AccordionContent = styled('div')({
   overflowY: 'hidden',
 
   [TABLET_BP_UP]: {
-    paddingTop: '2rem',
-    paddingBottom: '2rem',
+    padding: '3rem',
+    '& p': {
+      margin: 0,
```
palmenhq, line 105: Just a question, shall we have some kind of margin-bottom here? Just to separate the paragraphs visually. Or is that intentional?

palmenhq (dismissed):
Amazing! Welcome to hedvig 👋

palmenhq:
@styrelius feel free to use the "Update branch" button to get it in sync with master :)

gustaveen:
Whoop whoop, first PR and what a great one! 🎉

src/blocks/AccordionBlock/AccordionBlock.tsx
gustaveen: True, I think we have some multi paragraph accordions

styrelius: Ah OK, but maybe not on the last `p` then? Also I realised that the spacing between the `AccordionContent` and the next `AccordionTitle` is smaller in Figma so maybe the titles/headers should only have `paddingBottom` instead of both top and bottom padding?

gustaveen: Ah good catch, no `margin-bottom` on the last one 🙌

gustaveen: I guess the titles need the same padding top and bottom to align properly when not expanded. So should we maybe decrease the `padding-bottom` on accordion content

gustaveen (approved):
Awesome Ellen! 🍾

palmenhq (approved):
Boom! 💥 Nice

PR #525, opened 2020-09-03
Chore/replace tslint with eslint

gustaveen:
Took a stab at adding eslint 🥳 Works beautifully with prettier. I have turned some rules off to silence some errors that we can look over together and see what rules we want to add/remove. As well as adding some other nice eslint plugins.


styrelius (approved):
Looks good to me! 👏

PR #527, opened 2020-09-07
Add missing parenthesis

nordberg:
Noticed a missing parenthesis on the website.


palmenhq (approved):
Hello! Thanks for your contribution 🙌 The edited data is only used for testing purposes and the "real" data lives in a separate service which is not exposed on GitHub. However, many thanks for highlighting this - I will make sure to have it properly fixed later today!

PR #528, opened 2020-09-07
Chore/import eslint rules

gustaveen:
Adding import eslint plugin and set up rules for sorted imports. 
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md


styrelius (commented):
Nice! 👏 Isn't it even nicer though to also have an empty line between import groups? The `'newlines-between': 'always'` rule?

gustaveen:
> Nice! 👏 Isn't it even nicer though to also have an empty line between import groups? The `'newlines-between': 'always'` rule?

I think I prefer without newline to keep the file shorter but I can go either way :) What about you @palmenhq

palmenhq (approved):
woho!

PR #529, opened 2020-09-07
Update some ie11 fixes

palmenhq:


src/server/page.tsx
```diff
@@ -87,7 +87,10 @@ const template = ({
     }
     <div id="react-root">${body}</div>
       <script nonce="${nonce}">
-      window.__INITIAL_STATE__ = ${JSON.stringify(initialState)};
+      window.__INITIAL_STATE__ = ${JSON.stringify(initialState).replace(
```
gustaveen, line 90: Ah! This fixes the issue with the hidden blank space that broke the page right?

gustaveen (approved):
🥳

src/server/page.tsx
palmenhq: Yes!

PR #533, opened 2020-09-09
Fetch redirects from Storyblok

gustaveen:
Move redirect management to Storyblok 

- [x] Fetch redirects from Storyblok

- [x] Fetch redirects on request

- [x] Redirect fetched paths but bypass regular request (can only get one of them working at the same time 🤷‍♂️)

### DIFF
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -2,6 +2,7 @@ import React from 'react'
 import { BlogPostsPage } from './pages/BlogPostsPage'
 import { BlogPostsTagPage } from './pages/BlogPostsTagPage'
 import { PageFork } from './pages/PageFork'
+import { getDatasourceEntries } from './server/utils/storyblok'
 
 export interface Route {
   path: string
@@ -237,3 +238,10 @@ export const redirects: ReadonlyArray<[string, string, number]> = [
     302,
   ],
 ]
+
+export const getRedirects = async () => {
+  return await Promise.all([
+    getDatasourceEntries('permanent-redirects'),
+    getDatasourceEntries('temporary-redirects'),
+  ])
+}
--- a/src/routes.ts
+++ b/src/routes.ts
@@ -2,7 +2,6 @@ import React from 'react'
 import { BlogPostsPage } from './pages/BlogPostsPage'
 import { BlogPostsTagPage } from './pages/BlogPostsTagPage'
 import { PageFork } from './pages/PageFork'
-import { getDatasourceEntries } from './server/utils/storyblok'
 
 export interface Route {
   path: string
@@ -238,10 +237,3 @@ export const redirects: ReadonlyArray<[string, string, number]> = [
     302,
   ],
 ]
-
-export const getRedirects = async () => {
-  return await Promise.all([
-    getDatasourceEntries('permanent-redirects'),
-    getDatasourceEntries('temporary-redirects'),
-  ])
-}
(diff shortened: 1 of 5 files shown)


src/routes.ts
```diff
@@ -237,3 +238,10 @@ export const redirects: ReadonlyArray<[string, string, number]> = [
     302,
   ],
 ]
+
+export const getRedirects = async () => {
+  return await Promise.all([
```
palmenhq, line 243: I don't think we need async/await here :)

src/server/entry.tsx
```diff
@@ -91,12 +92,21 @@ router.get('/:locale(se|se-en|no|no-en)/referrals/:code', async (ctx) => {
...
-redirects.forEach(([source, target, code]) => {
-  router.get(source, (ctx) => {
-    ctx.status = code
-    ctx.redirect(target)
+getRedirects()
+  .then(async (data) => {
+    appLogger.info(`redirect data ${JSON.stringify(data)}`)
+    data.map((datasource) => {
+      datasource.datasource_entries.map((redirect: DatasourceEntry, index) => {
+        router.get(redirect.name, (ctx) => {
```
palmenhq, line 100: While this is clever it'll require us to restart the servers each time we change a redirect, ultimately leaving the ability to edit the data to developers anyways. I think rather we should write our own middleware that just gets the redirects and checks the path "run time" instead if "start time". Since we're using `cachedGet` anyways it should be fine to shoot that request every page load because it's stored in redis and redis is XtrEmelY speedy

gustaveen:
Can't get external redirects working with the current solution...

Screenshot 2020-09-11 at 17 44 23

src/server/middlewares/redirects.ts
```diff
@@ -33,3 +34,26 @@ export const startPageRedirect: IMiddleware<object> = async (ctx) => {
...
+    }
+  })
+
+  temporaryRedirects.datasource_entries.map((redirect) => {
+    if (redirect.name === ctx.originalUrl) {
+      ctx.redirect(redirect.value)
+      ctx.status = 302
+      return
+    }
+  })
```
palmenhq, line 56: the returns here will just return inside the `map`, not the outer function. Rather than redirecting inside the map, try using for example `find` to locate if there's any redirects matching. I.e.

```ts
const redirect = temporaryRedirects.datasource_entries.find(redirect => redirect.name === ctx.path)
if (redirect) {
  // Perform redirect
}
```

Also i'd suggest we'll use `path` instead of `originalUrl` because if someone gives a random query string we risk losing it. I think we should append the query to the redirect or something rather?

gustaveen: Haha how could I miss that, thanks 🤦‍♂️

Ah, that makes sense to add possible query string, added now 👍

src/server/middlewares/redirects.ts
```diff
@@ -33,3 +34,32 @@ export const startPageRedirect: IMiddleware<object> = async (ctx) => {
...
+  const temporaryRedirect = temporaryRedirects.datasource_entries.find(
+    (redirect) => redirect.name === ctx.path,
+  )
+
+  if (temporaryRedirect) {
+    const queryStringMaybe = ctx.querystring ? '?' + ctx.querystring : ''
+    ctx.redirect(`${temporaryRedirect.value}${queryStringMaybe}`)
+    ctx.status = 302
+    return
+  }
```
palmenhq, line 62: We do have some duplicate code here, maybe we could generify it a bit just to not repeat ourselves?

palmenhq (commented):
Startin' to look awesome! Shall we kill some of the hard-coded redirects before we merge though?

src/server/middlewares/redirects.ts
gustaveen: ofc! 🙌

gustaveen:
I have imported all the redirects in Storyblok now as well

palmenhq (approved):
great!

gustaveen:
Will test in staging after merge

PR #534, opened 2020-09-10
fix routes

AlexSJ1990:


palmenhq (approved):
This may be the last time you're doing this :D just saying #533

PR #535, opened 2020-09-10
Fix / Change how images and header scale

styrelius:
### This PR entails:
1) Changing `SITE_MAX_WIDTH` to `{max-width: 100%}` in order for content in header to stay pushed to the edges even on larger screens, and for images on the home page to scale up and stay at the edge on larger screens.

2) Change width of `IconWrapper` in `FooterBlock` by wrapping it with a `<ContentWrapper brandPivot>` since it was affected by the changes of `SITE_MAX_WIDTH` variable and got positioned weirdly.

---

### Ticket reference:
https://www.notion.so/hedviginsurance/Scale-a7ea6180798242d2af737ed0a6a71d54

---

### Screenshots:

**New design:**
20-09-10-new-img-scaling
20-09-10-new-footer-icon

**Old design:**
20-09-10-old-img-scaling
20-09-10-old-footer-icon


src/components/blockHelpers.tsx
```diff
@@ -24,10 +24,8 @@ export const LAPTOP_BP_UP = '@media (min-width: 1024px)'
 export const GIANT_BP_UP = '@media (min-width: 1700px)'
 
 export const SITE_MAX_WIDTH = {
-  maxWidth: 1384,
-  [GIANT_BP_UP]: {
-    maxWidth: 1500, // TODO: Update this value when we have design for large screens
```
palmenhq, line 29: lol  😅

src/components/blockHelpers.tsx
```diff
@@ -24,10 +24,8 @@ export const LAPTOP_BP_UP = '@media (min-width: 1024px)'
...
-  },
+  // TODO: Decide whether it's a good idea to have this kind of global styling variables here (or at all)
+  maxWidth: '100%',
```
palmenhq, line 28: I don't understand what this todo is about? But is it intentional to change the whole site to bet 100% width always? I feel like some blocks might not look amazing or?

palmenhq (commented):
Cool :) Just a question really

src/components/blockHelpers.tsx
styrelius: Oh, I added the `TODO` since I think that we should look through how these styling variables are used and which of them should be global in this way etc. The way it is now it was kind of tricky just to make the changes requested in the ticket. But I could remove it.
And yes, it is intentional with `maxWidth: 100%`. As it says in the ticket the max width was to be removed from header and the container for home page images and I set it to 100% just to make sure stuff wouldn't accidentally be wider than their parents. As for whether it looks good or not, I don't have a strong opinion on that. But the only content affected are content within a `ContentWrapper` with the prop `fullWidth` and not many blocks use that aside from the blocks with images and text on the home page and the header. Also this is only a thing for larger screens and not many "regular" persons have very large screens I think? 🤔
OK sorry for this lengthy comment but I guess I didn't do such a great job explaining in the PR description.

src/components/blockHelpers.tsx
styrelius: 🙃

src/components/blockHelpers.tsx
styrelius: Removed the `TODO`.

palmenhq (approved):
Woho! nice 🥳

gustaveen:
Awesome! @styrelius 🙌

PR #537, opened 2020-09-18
Remove all hardcoded routes

gustaveen:


styrelius:
Hm, what's going on here? 🤔

gustaveen:
> Hm, what's going on here? 🤔

Haha valid question. It turns out I have been using an old gitconfig with my old work mail so all my commits are attached to that email 🤦‍♂️ so I'm trying to attached the correct one now, i.e. that will change all history. But I will delete this branch and do it in a new one instead

styrelius:
> Haha valid question. It turns out I have been using an old gitconfig with my old work mail so all my commits are attached to that email 🤦‍♂️ so I'm trying to attached the correct one now, i.e. that will change all history. But I will delete this branch and do it in a new one instead

Ah, ouch! 😅

PR #539, opened 2020-09-21
Feature / add DK structure slash enable easier adding of new markets/locales 

styrelius:
### What?
This PR suggests new ways of adding and accessing values connected to a specific locale, like `hrefLang` and market names for example:
1) In `locales.ts` there is a big object containing one object for every possible locale with all values that differ between markets/locales. The one object with the current locale is then saved as context (through `constate`) and can be accessed throughout the repo. The other, not current, locales are used in order to get the right values for the `LanguagePicker` and `MarketPicker`.
2) There is also one addition to Storyblok `GlobalStory`: `four_oh_four_title`. This is used in the `FourOhFourBlock`.

In addition to this a few UX tweaks are made in `LanguagePicker` and `MarketPicker` - like disabling clicks on the current active language link in `LanguagePicker` and setting value of the `MarketSelect` in `MarketPicker` to the current market since the value of the dropdown isn't used to redirect and there was this strange behaviour where it looked like you'd change market but there was no redirect to that market specific version of the site. 

### Why?
The previous solution to locale specific data was changing several hard coded strings in conditional statements in a bunch of different files. This doesn't really scale so well and the changes suggested here will make it a lot easier when going to a new market since there will only be one place where stuff needs to be added in conjunction with creating a new dimension in Storyblok and this one place is the `locales.ts` file.

### GIFs
2020-09-23-languagepicker

2020-09-23-marketpicker-no

2020-09-23-marketpicker-da


src/components/containers/ContextContainer.tsx
```diff
@@ -1,8 +1,9 @@
-import { Container } from 'constate'
 import React from 'react'
+import { Container } from 'constate'
+import { LocaleData } from 'utils/locales'
 
-interface Context {
-  lang: string
+type Context = {
+  currentLocale: LocaleData
 }
```
palmenhq, line 7: Maybe we should use `interface` still?

palmenhq (dismissed):
Really REALLY really well done!
Just a couple of optional naming suggestions

src/blocks/FooterBlock/MarketPicker/MarketPicker.tsx
```diff
@@ -13,31 +19,47 @@ const MarketSelect = styled(Select)`
...
+  return locales.filter((locale) => locale.marketName !== currentMarket)
+}
+
 export const MarketPicker: React.FC<MarketPickerProps> = ({
-  currentLanguage,
+  currentLocale,
 }) => {
-  const marketLabel = getMarketLabel(currentLanguage)
-  const activeMarket = getCurrentMarket(currentLanguage)
+  const currentMarket = currentLocale.marketName
```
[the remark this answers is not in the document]
styrelius: Yup, that's better, will change that! 👍

src/components/containers/ContextContainer.tsx
styrelius: I'm going to leave the `type` alias since we agreed that we might transition to using `type` over `interface` for component props and state, and that it's fine that both are used in the codebase for now. 😇

palmenhq (dismissed):

src/server/middlewares/redirects.ts
```diff
@@ -26,20 +27,16 @@ export const forceHost = ({
...
   const queryStringMaybe = ctx.querystring ? '?' + ctx.querystring : ''
 
-  if (country === 'NO') {
-    ctx.redirect(`/no${queryStringMaybe}`)
-    return
+  if (Object.keys(locales).includes(countryLabel)) {
+    ctx.redirect(`/${countryLabel}${queryStringMaybe}`)
+  } else {
+    ctx.redirect(`/${fallbackCountryLabel}${queryStringMaybe}`)
   }
```
gustaveen, line 39: Maybe just have a early return? :) 
```suggestion
  if (Object.keys(locales).includes(countryLabel)) {
    ctx.redirect(`/${countryLabel}${queryStringMaybe}`)
    return
  }
  
  ctx.redirect(`/${fallbackCountryLabel}${queryStringMaybe}`)

```

gustaveen:
Amazing job with this Ellen! Good to release IMO! 🥳

src/server/middlewares/redirects.ts
styrelius: Sure, I don't have very strong opinions on using the `else` 🙃

gustaveen (approved):
Let's 🚢   it!

PR #540, opened 2020-09-22
Add query string to favicons to forces clients to fetch the newest one

gustaveen:
## Why
The old logo was shown as favicon. When removing the old favioncs we reverted back to old path. However, clients heavily cache favicons so the only way to force an update is to change the path. And adding a query string feels like the least cumbersome. 

IMG_3063

### DIFF
--- a/src/server/utils/favicons.ts
+++ b/src/server/utils/favicons.ts
@@ -1,19 +1,19 @@
 export const favicons = `
-<link rel="apple-touch-icon" sizes="57x57" href="/assets-next/favicons/apple-icon-57x57.png">
-<link rel="apple-touch-icon" sizes="60x60" href="/assets-next/favicons/apple-icon-60x60.png">
-<link rel="apple-touch-icon" sizes="72x72" href="/assets-next/favicons/apple-icon-72x72.png">
-<link rel="apple-touch-icon" sizes="76x76" href="/assets-next/favicons/apple-icon-76x76.png">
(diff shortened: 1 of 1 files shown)


styrelius:
Wouldn't it be possible to change the path to to `/assets` instead of `/assets-next` since that's what the folder is called? I know there's some mapping thingy that makes the `/assets-next` paths point to the /assets folder so I guess that means we would have to change that too, as well as changing to `/assets` in a few other files . Is that too much work? Or maybe there are other problems that this would entail that I don't know about 🤔

gustaveen:
> Wouldn't it be possible to change the path to to `/assets` instead of `/assets-next` since that's what the folder is called? I know there's some mapping thingy that makes the `/assets-next` paths point to the /assets folder so I guess that means we would have to change that too, as well as changing to `/assets` in a few other files . Is that too much work? Or maybe there are other problems that this would entail that I don't know about 🤔

I'm not sure if there's a reason for the name `/assets-next` that we mount in `configureAssets`, do you have any insight @palmenhq Anyhow, changing the name for the assets path this wouldn't fix the root problem in the long run

styrelius:
> I'm not sure if there's a reason for the name `/assets-next` that we mount in `configureAssets`, do you have any insight @palmenhq Anyhow, changing the name for the assets path this wouldn't fix the root problem in the long run

True.

styrelius (approved):

PR #543, opened 2020-10-12
[HVG-144] Fix / cookie consent component

styrelius:
### What?
Get text content for `CookieConsent.tsx` from Storyblok global story instead of having it hard coded in the component:
1) Add a `cookie_consent_message` to Storyblok `components.json` file and add a `cookie_consent_message` property of type `MarkdownHtmlComponent` to the `GlobalStory` interface in `StoryContainer.tsx`.
2) Replace hard coded text and conditional render from `CookieConsent` component and replace it with a `<GlobalStoryContainer>` in order to render the `cookie_consent_message` html as `dangerouslySetInnerHTML`.
3) Add a `cookie_consent_message` object to `globalStoryMock` in `storybook.ts` since TypeScript was screaming about it.

Finally make some minor styling fixes and change the x symbol in `CloseButton` in `CookieConsent` to one that isn't positioned weirdly in the button. It's also bigger and I hope that's fine :)

### Why?
The cookie consent/information element did not display copy in the right language, since the hard coded locale strings that was checked to conditionally render the right text (the text was also hard coded and it really shouldn't be) wasn't right.

_Referenced ticket [here](https://hedvig.atlassian.net/browse/HVG-144)_

### Screenshots
_/se:_
2020-10-12-cookie-consent-smaller-x-se

_/no-en:_
2020-10-12-cookie-consent-smaller-x-no-en

### DIFF
--- a/src/components/CookieConsent.tsx
+++ b/src/components/CookieConsent.tsx
@@ -2,10 +2,7 @@ import styled from '@emotion/styled'
 import { colorsV2 } from '@hedviginsurance/brand'
 import Cookies from 'js-cookie'
 import React, { useState, useEffect } from 'react'
-import {
-  CONTENT_GUTTER,
-  CONTENT_MAX_WIDTH_DEPRECATED,
-} from 'components/blockHelpers'
+import { CONTENT_GUTTER, CONTENT_MAX_WIDTH } from 'components/blockHelpers'
 import { GlobalStoryContainer } from 'storyblok/StoryContainer'
 
 const OuterWrapper = styled('div')<{ visible: boolean; closing: boolean }>(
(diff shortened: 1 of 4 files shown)


gustaveen:
Perfect! 🥳 Could we maybe add `font-size: 1.25rem` to `CloseButton` to make the cross a bit smaller :)

styrelius:
> Perfect! 🥳 Could we maybe add `font-size: 1.25rem` to `CloseButton` to make the cross a bit smaller :)

Haha I kind of like the exaggerated size but sure, I'll change it! Maybe it looks a bit stupid 😄

gustaveen (approved):
🍪👌

gustaveen:
We have to remember to add the text as soon as it's released though :)

styrelius:
> We have to remember to add the text as soon as it's released though :)

Yes that is very true

PR #545, opened 2020-10-16
Chore/add react hooks eslint plugin

gustaveen:
https://reactjs.org/docs/hooks-rules.html#eslint-plugin


palmenhq (approved):
NICE!

styrelius:
Awesome! ✨

src/blocks/HeaderBlockBrandPivot/index.tsx
```diff
@@ -236,7 +236,7 @@ export const Header: React.FC<{ story: GlobalStory } & HeaderBlockProps> = (
     minimalColorComponentColors | undefined
   >(props.cta_color?.color)
 
-  const updateHeader = () => {
+  const updateHeader = useCallback(() => {
```
styrelius, line 239: Just curious about the `useCallback` here since I've never really understood when you should use it! What's the reason for using it here? 🤓

gustaveen: `updateHeader` in this case is a dependency of the useEffect hook below it. useCallback memoize the function, preventing it from creating a new instance of the updateHeader function on every render which can lead to performance issues.

PR #546, opened 2020-10-20
Feature/dk perils

gustaveen:


palmenhq (commented):
Awesome! BUT - I think we need to white list the variable also in the [server webpack config](https://github.com/HedvigInsurance/web-next/blob/master/webpack/webpack.config.server.js#L8) right?

gustaveen:
> Awesome! BUT - I think we need to white list the variable also in the [server webpack config](https://github.com/HedvigInsurance/web-next/blob/master/webpack/webpack.config.server.js#L8) right?

Ah ofc, thank you mister! 🙌

I have added the variables in Heroku already

gustaveen:
Should I add it to dev webpack config as well to make it work locally? https://github.com/HedvigInsurance/web-next/blob/master/webpack/webpack.config.client.development.js#L36

palmenhq:
@gustaveen oh.. Actually good point. I don't think this works at all. Sorry must've missed that 😓  Since the code is run in the browser it doesn't have access to the environment variables on the server and hence `process.env.GIRAFFE_ENDPOINT` is just undefined al the time in the browser. Rather we should "forward" the environment variables via the `window` object or something. Like so: https://github.com/HedvigInsurance/web-next/blob/master/src/server/page.tsx#L91

gustaveen:
Ah yes, silly me! I guess that way is the easiest. Or do it in the webpack config. WDYT? @palmenhq 
```
const env = dotenv.config().parsed;
const giraffeEndpoint = JSON.stringify(env['GIRAFFE_ENDPOINT'])

module.exports = webpackConfig({
  ...
  plugins: [
    new webpack.HotModuleReplacementPlugin(),
    new webpack.DefinePlugin({
      'process.env': {
        NODE_ENV: JSON.stringify('development'),
        GIRAFFE_ENDPOINT: giraffeEndpoint,
      },
    }),
  ],
```

palmenhq:
Webpack config won't work because we use the same build for dev and production. So if we want to change them between environments we need to do it via the server.

palmenhq (approved):
great!

PR #547, opened 2020-11-10
Recognise locale correctly for hreflang

palmenhq:
Previously we accidentally didn't pass the correct locale to get the function that gets hreflang for a page alternative, meaning we fell back to `se` for every language.


vonElfvin (approved):

palmenhq:
Travis currently:



I'm just gonna merge this anyways

PR #548, opened 2020-11-12
Fix potential SEO bug

gustaveen:
Don't nest children inside H1. According to this report SEO is punished since server side rendering puts a style tag right before an element (crowding the H1). I'm not sure that's how google interprets it but better safe than sorry ey :D 

Screen Shot 2020-11-11 at 12 57 33

[HVG-236]

[HVG-236]: https://hedvig.atlassian.net/browse/HVG-236


styrelius (approved):
👍 👍 🦺 ⛑️ 🧯🥽

PR #549, opened 2020-11-13
Remove the background animation on all blocks

palmenhq:
To improve the first contentful paint speed and imo make loading nicer?
For me, running with and without animation locally, lighthouse performance is bumped from 57 to a good 66

So instead of this:

ezgif com-video-to-gif(3)

We do this:
ezgif com-video-to-gif(4)

### DIFF
--- a/src/blocks/HeaderBlockBrandPivot/index.tsx
+++ b/src/blocks/HeaderBlockBrandPivot/index.tsx
@@ -234,7 +234,11 @@ export const Header: React.FC<{ story: GlobalStory } & HeaderBlockProps> = (
   const [isBelowThreshold, setIsBelowThreshold] = useState<boolean>(false)
   const [buttonColor, setButtonColor] = useState<
     minimalColorComponentColors | undefined
-  >(props.cta_color?.color)
+  >(
+    props.inverse_colors && props.is_transparent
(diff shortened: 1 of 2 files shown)


gustaveen:
Ah so we get punished for delaying paint. Out of curiosity, would we get better scores if we remove the animation delay? But I'm good with this as a starter since I'm not sure if we want to fade in that part at all. 🙌

gustaveen (dismissed):

palmenhq:
@gustaveen probably we'd improve it but not as much as removing it. IMO it look better without the animation tho 😅  Do you think we should run this through someone else or should i just merge?

gustaveen:
No I think we can merge :) We have been talking about looking over the fade in animations anyway, and I agree that at least the hero can do with out it to get a snappier feel 🙌

styrelius (dismissed):
I agree, it's nicer without the initial white screen 👌

styrelius:
Just one thing I noticed: it feels a bit strange now though that the CTA changes color from black to white?

palmenhq:
@styrelius Nice catch! I've amended the pr

styrelius (approved):
Nice! 🎨

PR #550, opened 2020-11-13
Security headers

palmenhq:
Let's put back some security headers but disable CSP. All these other ones should be fine to have in here and _should_ not affect the tracking scripts. Rather, they should make us more likable to google.


gustaveen:
Was unclear to me what just adding `koa-helmet` does, but for reference it adds this by default:
https://github.com/venables/koa-helmet#usage
```
// This...
app.use(helmet());

// ...is equivalent to this:
app.use(helmet.contentSecurityPolicy());
app.use(helmet.dnsPrefetchControl());
app.use(helmet.expectCt());
app.use(helmet.frameguard());
app.use(helmet.hidePoweredBy());
app.use(helmet.hsts());
app.use(helmet.ieNoOpen());
app.use(helmet.noSniff());
app.use(helmet.permittedCrossDomainPolicies());
app.use(helmet.referrerPolicy());
app.use(helmet.xssFilter());
```

gustaveen (approved):
Yes please! 🔐

PR #551, opened 2020-11-13
Use 301 for start page redirect

palmenhq:
Because apparently google doesn't like 302s at all 🙄

### DIFF
--- a/src/server/middlewares/redirects.ts
+++ b/src/server/middlewares/redirects.ts
@@ -32,11 +32,14 @@ export const startPageRedirect: IMiddleware<object> = async (ctx) => {
     lookupCountry(actualIp)?.toLowerCase() || fallbackCountryLabel
   const queryStringMaybe = ctx.querystring ? '?' + ctx.querystring : ''
 
+  ctx.status = 301
+
   if (Object.keys(locales).includes(countryLabel)) {
-    return ctx.redirect(`/${countryLabel}${queryStringMaybe}`)
+    ctx.set('Location', `/${countryLabel}${queryStringMaybe}`)
+    return
   }
 
-  ctx.redirect(`/${fallbackCountryLabel}${queryStringMaybe}`)
+  ctx.set('Location', `/${fallbackCountryLabel}${queryStringMaybe}`)
 }
 
 const triggerRedirect = (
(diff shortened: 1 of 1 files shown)


src/server/middlewares/redirects.ts
```diff
@@ -32,11 +32,14 @@ export const startPageRedirect: IMiddleware<object> = async (ctx) => {
...
+  ctx.status = 301
+
   if (Object.keys(locales).includes(countryLabel)) {
-    return ctx.redirect(`/${countryLabel}${queryStringMaybe}`)
+    ctx.set('Location', `/${countryLabel}${queryStringMaybe}`)
+    return
   }
 
-  ctx.redirect(`/${fallbackCountryLabel}${queryStringMaybe}`)
+  ctx.set('Location', `/${fallbackCountryLabel}${queryStringMaybe}`)
```
gustaveen, line 42: How is `ctx.set` different from `ctx.redirect`? I thought adding `ctx.status = 301` would be sufficient? @palmenhq

palmenhq: i thought so also but it seems `ctx.redirect` overwrites with a `302`. I guess we could use `ctx.status` after `ctx.redirect` but i'm quite confident they will have the exact same effect or so i'm assuming, since setting the `Location` header + `301`/`302`(/`307`) is how you perform redirects with HTTP

styrelius (dismissed):
Nice! ✨  Does this mean that with the `ctx.redirect` you will automatically get the status `302`?

**Edit:** OK that's the same question that Gustav had haha! _Nevermind_

src/server/middlewares/redirects.ts
styrelius: image
☝️ From the koa docs though.

palmenhq: Hm strange, i must've been to eager while testing or something because i'm soo sure i got a 302. But maybe the server hadn't restarted yet or something. I can change it back and see what happens

palmenhq: Yup, it worked1

styrelius: Cool! 🆒 🤠

styrelius (approved):
🤖

PR #554, opened 2020-11-17
Replace https://www.hedvig.com with PUBLIC_HOST & fix frameguard (ie storyblok preview)

palmenhq:
This should only affect dev/staging environments, but basically ensures no (or at least fewer) links that are hard-coded to hedvig.com actually link to hedvig.com in the specific environment. If you're in the staging environment and press a cta you expect to reman in the staging environment right? This should ensure that behaviour to some extent, but can't handle if we link somewhere "directly" in copy, for example. But it's a first step at least 👍 

Extra 🌮 : Disable frameguard, since storyblok depends on iframes for its visual editor.

### DIFF
--- a/src/utils/storyblok.ts
+++ b/src/utils/storyblok.ts
@@ -1,9 +1,16 @@
 import { LinkComponent } from '../storyblok/StoryContainer'
 
-export const getStoryblokLinkUrl = (link: LinkComponent) =>
-  link.linktype !== 'story' || /^\//.test(link.cached_url)
-    ? link.cached_url
-    : `/${link.cached_url}`
+export const getStoryblokLinkUrl = (link: LinkComponent) => {
+  const cachedLink =
+    link.linktype !== 'story' || /^\//.test(link.cached_url)
+      ? link.cached_url
+      : `/${link.cached_url}`
+  const publicHost = getPublicHost()
+  return publicHost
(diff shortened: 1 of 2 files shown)


src/utils/storyblok.ts
```diff
@@ -1,9 +1,16 @@
...
-    ? link.cached_url
-    : `/${link.cached_url}`
+export const getStoryblokLinkUrl = (link: LinkComponent) => {
+  const cachedLink =
+    link.linktype !== 'story' || /^\//.test(link.cached_url)
+      ? link.cached_url
+      : `/${link.cached_url}`
+  const publicHost = getPublicHost()
+  return publicHost
+    ? cachedLink.replace('https://www.hedvig.com', publicHost)
```
gustaveen, line 10: Should we skip the replace if it's in prod?

gustaveen (approved):
Sounds reasonable! 🙌

src/utils/storyblok.ts
palmenhq: I was thinking about it but we'll substitute `https://www.hedvig.com` with `https://www.hedvig.com` and i think it dosen't really matter right? Maybe it'd be sliiiiightly more efficient from a performance perspecitve but i don't think it'd be in any way noticable

PR #555, opened 2020-11-18
Chore / Add pull request template

styrelius:
## What?

Add a template for pull request descriptions.

<br>

## Why?

We agreed it's a good thing and already implemented it in the web-onboarding repo 🤠

_Referenced ticket: [HVG-444]_
<br>

[HVG-444]: https://hedvig.atlassian.net/browse/HVG-444

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,20 @@
+<!-- 
+If there is one, add the ID of the Jira issue to the Pull Request title.
+E.g. "[HVG-123] Make the Pull Request template great again" 
+-->
+
+## What?
+
+<!-- What changes are made? If there are many significant changes, a list might be a good format. -->
+
+
+## Why?
+
+<!-- Why are these changes made? -->
+
+
+<!-- If there is one, add a link to the Jira issue below. -->
+_Referenced ticket [here]()_
+
+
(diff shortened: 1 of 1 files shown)


gustaveen:
Sorry if we already discussed this :D But I wonder if we need to have both the ticket number in the title and comment? Since it links automatically if you write it in the comment [HVG-236]

styrelius:
> Sorry if we already discussed this :D But I wonder if we need to have both the ticket number in the title and comment? Since it links automatically if you write it in the comment, if you write it like this `[HVG-236]` [HVG-236](https://hedvig.atlassian.net/browse/HVG-236)

Yes we discussed it in the PR in web-onboarding, but I'm not sure I get it - how is it automatically linked you mean? 

What I would like is
1) to have the issue ID in the title for Jira to pick it up, plus it's nice that you can see just from the title what issue it's regarding
2) a direct link to the issue so you don't have to go to Jira and search for it in order to find it

If there's a better way to do this than the one I suggest here I'm all for it! 😀

styrelius:
OK so you mean by "linked" that Jira picks it up and knows there a PR if you add the id like that in the description? So the same result as putting the id in the title? Then I still think it's nice to have both, for clarity 🙃

gustaveen (approved):
🥳

palmenhq (approved):
Awesome!

PR #556, opened 2020-11-23
Adjust Header height

gustaveen:
## What?

Adjust header layout to be consistent between market-web, onboarding and embark

## Why?

Because it removes ui glitches in the Header :)


styrelius (approved):
🆒

PR #558, opened 2020-12-01
Chore / Update packages

styrelius:
## What?

1) Upgrade a bunch of packages with `yarn upgrade`.

2) Remove `@hedviginsurance/cli` and `@hedviginsurance/react-swipeable-views` packages since we don't use them anymore.

3) Install the latest version of `@sentry/node` to fix some vulnerability of high severity connected to older versions of that package.

4) Remove the `sentryEnvironment` fallback `'dev'` in `sentry-server.ts` in order for my local `SENTRY_ENVIRONMENT` env variable to be picked up. Otherwise the environment will be "dev" in Sentry when you test the error logging locally.

5) Format a few files that caused Prettier linting errors after package updates.

## Why?

🧠 

~~_Referenced ticket: []_~~

[HVG-123]: https://hedvig.atlassian.net/browse/HVG-123

### DIFF
--- a/src/server/config/sentry.ts
+++ b/src/server/config/sentry.ts
@@ -1,5 +1,5 @@
 const sentryDsn = process.env.SENTRY_DSN
-const sentryEnvironment = process.env.SENTRY_ENVIRONMENT || 'dev'
+const sentryEnvironment = process.env.SENTRY_ENVIRONMENT
 
 export const sentryConfig = () => ({
   dsn: sentryDsn,
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "@emotion/styled": "^10.0.23",
     "@hedviginsurance/brand": "^4.0.3",
     "@segment/utm-params": "^2.0.0",
-    "@sentry/node": "^4.3.0",
+    "@sentry/node": "^5.27.6",
     "@types/koa-basic-auth": "^2.0.3",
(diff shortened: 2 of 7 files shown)


src/server/config/sentry.ts
```diff
@@ -1,5 +1,5 @@
 const sentryDsn = process.env.SENTRY_DSN
-const sentryEnvironment = process.env.SENTRY_ENVIRONMENT || 'dev'
```
virrios, line 2: Even if your env is set, does it fallback?
Don't we want to keep the fallback if we don't have it set?

virrios (commented):
Nice work!! 🙌 
just have one question

src/server/config/sentry.ts
styrelius: Yep, the env variable is just ignored if we have the fallback. Since it takes longer to get the env variable I guess - it's undefined for a fraction of a moment so then `sentryEnvironment` is set to `'dev'` and never change, at least locally. I mean, it works in production.
And correct me if I'm wrong - I don't think we need the fallback since the error would be logged to Sentry anyway - it will just not have a specified environment. And that is a more correct IMO than if it's just set to "dev" even if the environment isn't actually "dev".

virrios: Nice! 👌

virrios (approved):

PR #559, opened 2020-12-07
Fix / YoutubeVideoBlock infinite loop of rendering

styrelius:
## What?

1. Fix the infinite loop of re-rendering in `YoutubeVideoBlock.tsx`, caused by a dependency to a `useEffect`:
What we want is for the `VideoFrame` (a styled `iframe`) to have the same size as the `OverLayImage`. After solving the problem by replacing the `setInterval` function checking the size of the `OverLayImage` with a listener on the `'resize'` event in the order to change the 'playerWidth' and 'playerHeight' states, I realised that this issue could actually also be fixed by just setting the `VideoFrame` height to `100%` . 🙃

2. Some CSS refactoring, making some of the rules a little less complicated and extracting some values used in more than one place to variables. As far as I have been able to see none of this has changed any behaviour or styling of the rendered elements.

3. Change name of `setPlayerLoaded` to `setIsPlayerLoaded` (nothing else changed with this, just the name).

**Please note:** The autoplay doesn't always work, but that is the case without these changes as well. At some point that should probably be looked into but now is not the time in my opinion, since we have more important things to fix.

## Why?

The entire page gets disabled (nothing happens when you click a link etc.) if this block is used the way it is constructed now, due to the infinite loop of re-rendering.

These changes will also make this component less complex.

_Referenced ticket: [HVG-627]_

## GIF
**The Youtube Video Block used at /se/about-us - The same look and behaviour as before, minus the infinite loop:**

2020-12-07-youtube-video-block

[HVG-627]: https://hedvig.atlassian.net/browse/HVG-627

### DIFF
--- a/src/blocks/YoutubeVideoBlock/YoutubeVideoBlock.tsx
+++ b/src/blocks/YoutubeVideoBlock/YoutubeVideoBlock.tsx
@@ -124,7 +124,7 @@ export const YoutubeVideoBlock: React.FunctionComponent<YoutubeVideoBlockProps> 
   index,
 }) => {
   const [isPlaying, setIsPlaying] = useState(false)
-  const [isPlayerLoaded, setPlayerLoaded] = useState(false)
+  const [isPlayerLoaded, setIsPlayerLoaded] = useState(false)
   const [isShowingOverlay, setIsShowingOverlay] = useState(true)
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


gustaveen (approved):
Nice, yay for CSS! 🥳 You can let Carl know that this is fixed when you release this, since he temporarily removed the video on the about us pages

PR #561, opened 2020-12-11
Add Trustpilot reviews slider

gustaveen:
## What?

Added Trustpilot reviews slider. It's an embedded iframe since we don't have the time to build something custom and we want reviews NOW. There's no capabilities to edit the widget in Storyblok right now, we might add this later. If you're on `/se` review will be in Swedish , all other will be in English since we barely have any Norwegian reviews yet. Embed code from here: https://trstp.lt/YH5N4nS_m

Mange & CO will be very happy if we can get this out today 🥳

## Why?

This will be used specifically on campaign pages but on some other pages as well. 

And no, it will not be right here on the startpage :) 

trutpilotreview

### DIFF
--- a/src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
+++ b/src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
@@ -1,0 +1,54 @@
+import React, { useEffect, useRef } from 'react'
+import styled from '@emotion/styled'
+import { colorsV3 } from '@hedviginsurance/brand'
+import { ContentWrapper } from 'components/blockHelpers'
+import { ContextContainer } from 'components/containers/ContextContainer'
+
+const TrustpilotWrapper = styled(ContentWrapper)`
+  && {
+    max-width: 1200px;
+  }
+`
+
+export const TrustpilotBlock: React.FC = () => {
+  const trustpilotRef = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
+    const castedWindow = window as any
+    if (castedWindow && castedWindow.Trustpilot) {
+      castedWindow.Trustpilot.loadFromElement(trustpilotRef.current, true)
+    }
+  }, [])
+
+  return (
+    <TrustpilotWrapper>
+      <ContextContainer>
+        {({ currentLocale }) => (
+          <div
+            ref={trustpilotRef}
+            className="trustpilot-widget"
+            data-locale={currentLocale.label === 'se' ? 'sv-SE' : 'en-US'}
+            data-template-id="54ad5defc6454f065c28af8b"
+            data-businessunit-id="5b62ebf41788620001d3c4ae"
+            data-style-height="240px"
+            data-style-width="100%"
+            data-theme="light"
+            data-stars="4,5"
+            data-review-languages={
+              currentLocale.htmlLang === 'sv' ? currentLocale.htmlLang : 'en'
+            }
+            data-text-color={colorsV3.gray900}
+          >
+            <a
+              href="https://www.trustpilot.com/review/www.hedvig.com"
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              Trustpilot
+            </a>
+          </div>
+        )}
+      </ContextContainer>
+    </TrustpilotWrapper>
+  )
+}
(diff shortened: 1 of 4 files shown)


src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }, [])
+
+  return (
+    <TrustpilotWrapper>
+      <ContextContainer>
+        {({ currentLocale }) => (
+          <div
+            ref={trustpilotRef}
+            className="trustpilot-widget"
+            data-locale={currentLocale.label === 'se' ? 'sv-SE' : 'en-US'}
```
gustaveen, line 30: Yes it's not beautiful, but it gets the job done for now :)

styrelius: Lol you knew I would comment on this 😄 
OK fine, let's do it like this for now (but I guess it's not _that_ big of a fix to add the `data-locale` as a property on the `locales` objects right?)

gustaveen: That would work for sure! 😃  But wouldn't it be confusing to add `en-US` to dk and no `locales` object, or maybe not? Since this will only be used in one file.

It would be nice IMO to have a hook later on to get `currentLocale` so we can keep the component a bit cleaner and we don't we don't have to use `<ContextContainer>`

src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
```diff
@@ -0,0 +1,54 @@
...
+import styled from '@emotion/styled'
+import { colorsV3 } from '@hedviginsurance/brand'
+import { ContentWrapper } from 'components/blockHelpers'
+import { ContextContainer } from 'components/containers/ContextContainer'
+
+const TrustpilotWrapper = styled(ContentWrapper)`
+  && {
+    max-width: 1200px;
+  }
+`
```
styrelius, line 11: I'm wondering if this is really necessary? If you don't pass any particular props to `ContentWrapper` you'll get this exact `max-width`. And if you still want to add this styling to be sure it's always there, even if someone change or remove the `CONTENT_MAX_WIDTH_DEPRECATED` variable in `blockHelpers.tsx" it's enough with this I think:
```suggestion
const TrustpilotWrapper = styled(ContentWrapper)`
  max-width: 1200px;
`
```

styrelius (commented):
Great! 🙌 Just one thing (in addition to the one you already anticipated 😄)

src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
styrelius: I don't think it would be that confusing, if the property is named so it's clear it's only used for TrustPilot. And I can see what you mean with that we only use it in one file, but we also have the `adtractionSrc` property on the `locales` which are also only used in one place. I think it's nice to just have all that data in one place and make the actual components a bit cleaner without too many `if` and ternaries. Because I can see a scenario pretty soon that we'll have more reviews in Norway (in Norwegian) and then we'd have to add that case too to the conditional.

But I totally agree we should do a remake to how we use context and get rid of the `ContextContainer` by utilising the `useContext` hook for sure! Let's add that do the infinite list of todo's! 😀

src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
gustaveen: That is correct! But it also gives me a `maxWidth: 1500` that looks way off for this particular component. Since I don't want to remove `CONTENT_MAX_WIDTH_DEPRECATED` in this PR I need to override that with `&&` 
```
export const CONTENT_MAX_WIDTH_DEPRECATED = {
  maxWidth: 1200,
  [GIANT_BP_UP]: {
    maxWidth: 1500,
  },
}
```

gustaveen: But I absolutely agree that the way we handle max-widths at the moment is not optimal 😄

styrelius: Ah, OK I see 👍

src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
gustaveen: Ok, fair enough. 👍  Is this a deal breaker now or can we make a new PR on monday since I have meetings the rest of the day and we want it out today 🙏

styrelius: OK, we can merge it and I can fix it in another branch later today or on Monday. 🤖

styrelius (approved):
⭐

PR #562, opened 2020-12-11
Fix/update locales object

gustaveen:
## What?

Update to locales object with Trustpilot locale and lang

## Why?

Removing conditionals, however this adds more repetition IMO but I'm good if you think this makes it better 😀


styrelius (commented):
Nice! 🙌  Some ideas how to address the issue with too much repetition below. And also, if you don't want to work with this I can fix it later.

PR #569, opened 2021-04-12
Geo redirect for Hedvig goods

palmenhq:
## What?

Use Geo redirect for Hedvig goods. Also make it easy to geo redirect any page.

## Why?

🌍

_Referenced ticket: []_

### DIFF
--- a/src/server/middlewares/redirects.ts
+++ b/src/server/middlewares/redirects.ts
@@ -5,7 +5,7 @@ import * as Koa from 'koa'
 import { lookupCountry } from 'server/utils/ip2location'
 import { getDatasourceEntries } from 'server/utils/storyblok'
 import { DatasourceEntry } from 'storyblok/StoryContainer'
-import { locales, fallbackLocale } from 'utils/locales'
+import { fallbackLocale, locales } from 'utils/locales'
 import { State } from './states'
 
 export const forceHost = ({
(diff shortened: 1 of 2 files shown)


styrelius (commented):
Nice! 🙌 🌟  I'm just thinking that we should maybe have the geo redirects for all pages/routes? Instead of explicitly adding it for specific pages?
Doesn't feel great to have hard coded routes in the code base like this, when routes are changes from Storyblok. For example one issue is that this route will probably be removed again pretty soon, when the Hedvig Goods project is over, and then we'll have to remember to remove it from the codebase as well.

palmenhq:
@styrelius in general i agree here, however i'm not sure how feasable that is as of now... How do we know which pages should be geo redirected and not? Because i deffo don't think we should geo redirect to a 404. Maybe we could check if the target exists in storyblok, but then we need to check for translated slugs and stuf.. it feels a bit cumbersome. Maybe we could overlook this solution if we see the need?

gustaveen:
Yeah it feels a bit much to fix this for all routes as of now. This is mainly for marketing purposes, but I agree that it's always a risk that it would be left laying around.

src/server/middlewares/redirects.ts
```diff
@@ -25,21 +25,21 @@ export const forceHost = ({
...
   const queryStringMaybe = ctx.querystring ? '?' + ctx.querystring : ''
 
   ctx.status = 301
 
   if (Object.keys(locales).includes(countryLabel)) {
-    ctx.redirect(`/${countryLabel}${queryStringMaybe}`)
-    return
+    ctx.redirect(`/${countryLabel}${path}${queryStringMaybe}`)
+  } else {
+    ctx.redirect(`/${fallbackCountryLabel}${path}${queryStringMaybe}`)
```
gustaveen, line 41: I kinda like not having an `else` but no biggie :)

gustaveen (approved):
Nicely done! 🌍 🙌

src/server/middlewares/redirects.ts
palmenhq: I used to but have changed my mind, imo `else`s are more verbose and obvious than early returns

PR #570, opened 2021-04-13
Remove mandatory title in image text block

gustaveen:
## What?

Remove mandatory title in image text block

## Why?

It used a lot without the need for a title


styrelius (approved):
🏃💨

PR #571, opened 2021-04-19
Add correct peril types for 🇩🇰

gustaveen:
## What?

Add DK contract types for Peril block 

Screenshot 2021-04-19 at 16 10 42

### In Storyblok
Screenshot 2021-04-19 at 16 13 19


styrelius (approved):
👍 👍

PR #575, opened 2021-05-17
Remove fill-style from `PerilItem`

robinandeer:
## What?

Remove fill-style from SVG icon path in `PerilItem`-component.

## Why?

SVG icons control their own color styles.


gustaveen (approved):
🙏

PR #577, opened 2021-05-18
Update image text block layout/alignment

robinandeer:
## What?

Align text in block with menu (left).
Push text content down to bottom for size XL.
Ensure consistent bottom padding.

## Why?

The look of this element no longer matches the intended designs in Figma.

Editors have copy-pasted styles into StoryBlok to force layout changes.

> Menu and Intercom chat bubble are handled in different PRs

> When deploying these changes we also need to update the "extra_styling" in StoryBlok to remove `padding-bottom. It is now handled by the component.

## Mobile: same layout

Screenshot 2021-05-18 at 16 29 31

## Tablet: before

Screenshot 2021-05-18 at 16 30 05

## Tablet: after

Screenshot 2021-05-18 at 16 30 17

## Desktop: before

Screenshot 2021-05-18 at 16 31 05

## Desktop: after

Screenshot 2021-05-18 at 16 30 57

## Large desktop: before

Screenshot 2021-05-18 at 16 31 42

## Large desktop: after

Screenshot 2021-05-18 at 16 31 33

### DIFF
--- a/src/blocks/ImageTextBlockBrandPivot/index.tsx
+++ b/src/blocks/ImageTextBlockBrandPivot/index.tsx
@@ -40,7 +40,7 @@ interface Animateable {
 const AlignableContentWrapper = styled(ContentWrapper)<{
   textPosition: TextPosition
 }>(({ textPosition }) => ({
-  padding: '0 1rem',
+  padding: '0 2rem',
   display: 'flex',
   position: 'relative',
   flexDirection:
@@ -81,7 +81,7 @@ const TextWrapper = styled('div')<{
   textAlign: textPosition === 'center' ? 'center' : 'left',
   width: '100%',
   ...(textPosition === 'left' && {
-    paddingLeft: isBlockFullWidth ? '4rem' : 0,
+    paddingLeft: 0,
     paddingRight: '4rem',
   }),
   ...(textPosition === 'right' && {
--- a/src/blocks/ImageTextBlockBrandPivot/index.tsx
+++ b/src/blocks/ImageTextBlockBrandPivot/index.tsx
@@ -39,8 +39,9 @@ interface Animateable {
 
 const AlignableContentWrapper = styled(ContentWrapper)<{
   textPosition: TextPosition
-}>(({ textPosition }) => ({
-  padding: '0 2rem',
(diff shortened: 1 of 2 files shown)


gustaveen:
Nice tinkering! It's a bit of a mess with the combination of manual styles 😅 So it's hard to accomplish without having to go and change the manual styles. The only thing I noticed is that the added padding pushes in the fullWidth image blocks. I know I got feedback earlier that we wanted the full bleed images to have minimal padding. But on the other hand int aligns better with the menu 👌 But let's check this with Mange 🙌

### Before
Padding around the image is 16px consistently 
Screenshot 2021-05-20 at 10 36 40

### After
Screenshot 2021-05-20 at 10 36 53

src/blocks/ImageTextBlockBrandPivot/index.tsx
```diff
@@ -40,7 +40,7 @@ interface Animateable {
 const AlignableContentWrapper = styled(ContentWrapper)<{
   textPosition: TextPosition
 }>(({ textPosition }) => ({
-  padding: '0 1rem',
+  padding: '0 2rem',
```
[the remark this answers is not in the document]
robinandeer: Right, that's a good point. I'll have a look at what we need to do to accommodate that

robinandeer:
Screenshot 2021-05-20 at 16 51 25

src/blocks/ImageTextBlockBrandPivot/index.tsx
```diff
@@ -317,6 +318,7 @@ export const ImageTextBlockBrandPivot: React.FunctionComponent<ImageTextBlockPro
         index={index}
         brandPivot
         fullWidth={full_width}
+        padding={background_type === 'video' ? 'relaxed' : 'minimal'}
```
styrelius, line 321: I think you know more about this @gustaveen but shouldn't the `'relaxed'` padding also be used if we use an image and not a video as background? Or am I missing something?

robinandeer: My idea was the change as little as possible from the current implementation. That's why I try to target only when the block is used with a video element.

PR #579, opened 2021-05-20
Localised Trustpilot reviews for reals this time

gustaveen:
## What?

Fetch Trustpilot review based on tags instead. Tags is added manually by Emil in Trustpilot. The tags we use represent the language they are written in `sv, no, da, en` i.e. maps directly to html lang. 
The `data-locale` is the boilerplate text from the widget like dates and such and is written in iso format, however the delimiter is different from our iso unfortunately, so I added an extra property to the `locales` object  

## Why?

Add possibility to show Trustpilot reviews in native languages across the site. Having handpicked reviews as we had before limited us to 15 reviews for all markets.

### Trustpilot widget

https://user-images.githubusercontent.com/6661511/118940231-c9da9600-b950-11eb-9dc4-7c79acbdcfc4.mov

### DIFF
--- a/src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
+++ b/src/blocks/TrustpilotBlock/TrustpilotBlock.tsx
@@ -27,16 +27,17 @@ export const TrustpilotBlock: React.FC = () => {
           <div
             ref={trustpilotRef}
             className="trustpilot-widget"
-            data-locale={currentLocale.label === 'se' ? 'sv-SE' : 'en-US'}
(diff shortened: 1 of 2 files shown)


styrelius (approved):
Nice! 🎂 🎂 🎂  Awesome job talking to Trustpilot about this and finding this solution, which is not the most fun work of course.

gustaveen:
Thank you! Feels good to let the Trustpilot support at least do some work for all the money they get from us 😄

PR #580, opened 2021-05-20
Support standard color for app buttons block

robinandeer:
## What?

Pass color option to app buttons to turn them black on white (standard) background.
Add type guard function to keep things type safe. Maybe a bit complicated solution but at least it's type safe - what do you say?

## Why?

App buttons are white on white (standard) background. We are not passing the color option to the buttons component.

## Screenshot: hover vs. not-hover

Screenshot 2021-05-20 at 15 02 26

### DIFF
--- a/src/blocks/AppButtonsBlock.tsx
+++ b/src/blocks/AppButtonsBlock.tsx
@@ -1,7 +1,23 @@
 import React from 'react'
 import { AppButtons } from 'components/AppButtons/AppButtons'
 import { ContentWrapper, SectionWrapper } from 'components/blockHelpers'
-import { BrandPivotBaseBlockProps } from './BaseBlockProps'
+import {
+  BrandPivotBaseBlockProps,
+  minimalColorComponentColors,
+} from './BaseBlockProps'
+
+type StandardColor = 'standard' | 'standard-inverse'
+
+const STANDARD_COLORS: Array<minimalColorComponentColors> = [
+  'standard',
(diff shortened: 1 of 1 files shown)


src/blocks/AppButtonsBlock.tsx
```diff
@@ -1,7 +1,23 @@
...
+const STANDARD_COLORS: Array<minimalColorComponentColors> = [
+  'standard',
+  'standard-inverse',
+]
+
+function isStandardColor(
+  color: minimalColorComponentColors,
+): color is StandardColor {
+  return STANDARD_COLORS.includes(color)
+}
```
gustaveen, line 20: Might be just me that's too used to arrow functions :D But we try to stick to that for consistency or WDYT?

gustaveen (commented):
Cool! I'm good with this solution but since you still can choose the options in Storyblok this might be confusing :)

src/blocks/AppButtonsBlock.tsx
robinandeer: Sure, I don't mind either way! #tabsVSspaces

robinandeer:
> Cool! I'm good with this solution but since you still can choose the options in Storyblok this might be confusing :)

I agree, but I guess we don't even intend to support all colors, no? Happy to take any suggestions!

gustaveen (approved):
🙌

PR #584, opened 2021-06-23
Add possibility to display more links in `InsuranceInfoBlock`

robinandeer:
## What?

- You can now add four links in total, instead of two
- The links are grouped in Storyblok schema for a better overview in editor

## Why?

Need to prepare for Danish contents + travel + accident bundle that requires 3 links.
We added an option for a 4th link to prepare for the future.

Screenshot 2021-06-23 at 15 36 59


src/blocks/InsuranceInfoBlock/InsuranceInfoBlock.tsx
```diff
@@ -113,101 +127,60 @@ export const InsuranceInfoBlock: React.FC<InsuranceInfoBlockProps> = ({
...
+    { description: data.value_6_description, value: data.value_6_value },
+    { description: data.value_7_description, value: data.value_7_value },
+  ].filter((item) => item.value)
+
+  const links: Links = [
+    { link: data.terms_link, text: data.terms_link_text },
+    { link: data.terms_link_2, text: data.terms_link_text_2 },
+    { link: data.terms_link_3, text: data.terms_link_text_3 },
+    { link: data.presale_info_link, text: data.presale_info_link_text },
+  ].filter((item) => item.text)
```
gustaveen, line 147: Neat! 🙌

gustaveen (approved):
Good job, works like a charm! 🎉

It hits me though that it would have been better when we built this block to use a link field :) And we could probably turn the values and links into repeatable fields. But it's always a bit of a hustle to migrate to a new structure so let's not try that now :)

PR #585, opened 2021-07-13
Serve robots.txt file

robinandeer:
## What?

Serve robots.txt file.
Add new static route to serve public files from root domain.

## Why?

Suggestion from SEO audit.
Make sure that google and other search engines don't crawl parts of the website they shouldn't.

_Referenced ticket: [TSEO-1]_

[TSEO-1]: https://hedvig.atlassian.net/browse/TSEO-1

### DIFF
--- a/assets/public/robots.txt
+++ b/assets/public/robots.txt
@@ -1,0 +1,13 @@
+User-agent: *
+Disallow: */old-*
+Disallow: */referrals*
+Disallow: */forever*
+Disallow: */new-member/*
+Disallow: */welcome?code=
+Disallow: */partner*
+Allow: */partner-program
+Disallow: */a-b-test*
+Disallow: */invite*
+Disallow: */lp/*
+
+Sitemap: https://www.hedvig.com/sitemap.xml
--- a/assets/public/robots.txt
+++ b/assets/public/robots.txt
@@ -1,16 +1,8 @@
 # robots.txt for https://www.hedvig.com/
 
 User-agent: *
-Disallow: */old-*
 Disallow: */referrals*
-Disallow: */forever*
 Disallow: */new-member/*
-Disallow: */welcome?code=
-# Block access to discount code pages
-Disallow: */partner*
-Allow: */partner/partner-program
-Disallow: */a-b-test*
-Disallow: */invite*
 Disallow: */lp/*
 
 Sitemap: https://www.hedvig.com/sitemap.xml
--- a/assets/public/robots.txt
+++ b/assets/public/robots.txt
@@ -1,11 +1,14 @@
+# robots.txt for https://www.hedvig.com/
+
 User-agent: *
 Disallow: */old-*
 Disallow: */referrals*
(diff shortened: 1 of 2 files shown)


assets/public/robots.txt
```diff
@@ -0,0 +1,13 @@
+User-agent: *
+Disallow: */old-*
+Disallow: */referrals*
+Disallow: */forever*
+Disallow: */new-member/*
+Disallow: */welcome?code=
+Disallow: */partner*
+Allow: */partner-program
```
gustaveen, line 8: Why do we need to specifically allow this path? 🙃 Aren't all other paths crawled by default?

gustaveen (dismissed):
Nothing bad with adding a robots.txt file, but did they mention that we need to add `noindex` as well? Since a `Disallow` won't guarantee that we block all crawlers :) https://developers.google.com/search/docs/advanced/robots/intro

assets/public/robots.txt
robinandeer: It's because it's under `/*/partner/partner-program/*`

robinandeer:
> Nothing bad with adding a robots.txt file, but did they mention that we need to add `noindex` as well? Since a `Disallow` won't guarantee that we block all crawlers :) [developers.google.com/search/docs/advanced/robots/intro](https://developers.google.com/search/docs/advanced/robots/intro)

They did not mention that but I might as well go over the pages in storyblok to make sure we include that tag.

robinandeer:
I've gone over all the pages in Storyblok to make sure they are also passing the robots="noindex" meta tag.

styrelius (commented):
Cool 🤖 

So the way I understand it we shouldn't disallow pages in this file that we don't want indexed, but _only_ use this `noindex` meta tag / HTML header - is that what we're doing now or are we both disallowing here _and_ using the `noindex`?
https://developers.google.com/search/docs/advanced/crawling/block-indexing

robinandeer:
Let's maybe rethink this a little.

Google is pretty clear that `noIndex` pages should not be excluded in `robots.txt`!

However, there are some sections of the website that should be excluded from search so I think we should keep them in `robots.txt`.

We should also serve a `robots.txt` with a reference to the sitemap.

styrelius:
OK maybe I'm misunderstanding you but if we want pages to be excluded from search / not indexed by search engines shouldn't we then _only_ add the `noindex` meta tag on those pages and _not_ list them with `Disallow` key in the `robots.txt`? Or is the reasoning for example that it's not possible to add the `noindex` meta tag to all the pages under `/new-member`?

Also, what is `*/lp/*`? 🤔

robinandeer:
lp = landing page -- I think it's pages linked directly from ads maybe? I guess we don't want organic traffic to those pages. But no idea if they are still in use.

As I understood, `robots.txt` is better for excluding whole sections of a site so that's why I kept `/new-member` in there.

styrelius (approved):
🤖 N.I.C.E

PR #588, opened 2021-07-20
Render market links to initial HTML payload

robinandeer:
## What?

Implement combobox that keeps dropdown content mounted.
Copy over styles to match existing Select component.
Remove unused components.

## Why?

We want Googlebot and other search engine crawlers to discover the market site links.
This requires them to be part of the HTML rendered on the server.
`react-select` doesn't support this, so the links are not discovered by the crawlers.

This PR implements a simple combobox that better supports SEO without sacrificing usability.

> I used a button to open the dropdown menu which means that is has a hover effect. I think it looks rather nice though.

_Referenced ticket: [TSEO-13]_

### Demo

https://user-images.githubusercontent.com/1220232/126323737-f883326b-27cb-4940-8d32-64026bee77ff.mov

#### SSR-generated links

Screenshot 2021-07-20 at 14 21 49

[TSEO-13]: https://hedvig.atlassian.net/browse/TSEO-13


robinandeer:
Reference for aria props:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html

src/blocks/FooterBlock/MarketPicker/MarketPicker.tsx
```diff
@@ -1,70 +1,133 @@
...
+  list-style: none;
+  height: 3rem;
+  padding: 0.75rem 1.5rem;
+`
+
+const MenuItemLink = styled.a`
+  text-decoration: none;
+  display: block;
+  height: 100%;
+
```
styrelius, line 66: Maybe put the padding on the `a` instead of on the `li` to make the entire area clickable? 
```suggestion
const MenuItem = styled.li`
  list-style: none;
  height: 3rem;
`

const MenuItemLink = styled.a`
  text-decoration: none;
  display: block;
  height: 100%;
  padding: 0.75rem 1.5rem;

```

styrelius: Also, should this really be an `a` and not a `Link` from React Router? Or is it because we want to add the `hrefLang` attribute?

styrelius (commented):
Great job!! 🥳

src/blocks/FooterBlock/MarketPicker/MarketPicker.tsx
```diff
@@ -1,70 +1,133 @@
...
+            aria-selected={market.marketName === currentMarketName}
+          >
+            <MenuItemLink
+              href={`/${market.marketLabel}`}
+              hrefLang={market.hrefLang}
+            >
+              {market.marketName}
+            </MenuItemLink>
+          </MenuItem>
+        ))}
```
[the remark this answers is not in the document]
robinandeer: Yeah I think that would be a nice addition. I just kept it the same to keep the PR focused :-)

robinandeer:
> Also, should this really be an a and not a Link from React Router? Or is it because we want to add the hrefLang attribute?

@styrelius To me, it feels accurate to navigate with a regular link instead of client side routing. It's almost like an external link to me, what do you say? It's also how it used to work so perhaps that's why I decided to keep it that way.

We do want to use `href lang`. I would've guessed `<Link>` also supports it though.

styrelius:
> @styrelius To me, it feels accurate to navigate with a regular link instead of client side routing. It's almost like an external link to me, what do you say? It's also how it used to work so perhaps that's why I decided to keep it that way.

We do want to use href lang. I would've guessed <Link> also supports it though.

@robinandeer Hm, I'm not sure. But yeah, maybe it's more in the "external link" category, especially since we always just link to the home page regardless of where on the site the user is. Let's keep it (for now at least)! 👍

styrelius (approved):
🙌 🇸🇪 🇳🇴 🇩🇰

PR #589, opened 2021-07-21
Filter out non-public pages from sitemap

robinandeer:
## Why?

We need a way to exclude Storyblok pages from the sitemap.

As far as I could tell, all pages have the public checkbox enabled. It seems only relevant to exclude the non-public pages from the sitemap so maybe that's all we need.

_Referenced ticket: [TSEO-19]_

[TSEO-19]: https://hedvig.atlassian.net/browse/TSEO-19

### DIFF
--- a/src/server/sitemap.ts
+++ b/src/server/sitemap.ts
@@ -60,6 +60,7 @@ const getSitemapLinks = async (logger: Logger): Promise<ReadonlyArray<Url>> => {
   const sitemapLinks = pages
     .filter(({ url }) => Boolean(url))
     .filter(({ page }) => page?.story?.content?.robots === 'index')
+    .filter(({ page }) => page?.story.content?.public === true)
     .map(({ url }) => ({
       url,
       priority: 0.7,


styrelius (commented):
I don't really understand what the purpose of the **Public** property is - shouldn't all pages that are published be public? Or what does it mean that they aren't/are public?

robinandeer:
Hmm that's a valid point. I'm not sure how the editor flow is supposed to work with Storyblok.

Do you think it would be possible to unpublish content that is no longer supposed to be visible on the website? Perhaps we can ask Emil?

palmenhq:
Morning! 👋 For some reason i got a notification about this, so i figured i might as well give my two cents! 

I think this is legacy from when field-level translations was enabled in storyblok. Swedish/english pages were treated as one, and you just translated individual fields. eventually it turned out we wanted content for swedish only, for example, and there was no way of hiding the english content unless you added a custom field that could be "translated" (as publish/draft state was the same for all languages).

But as nowadays as folder-level translation is used (i assume still?) it's no longer needed i guess, especially since the filtering of this property was removed in PR #539 ([here](https://github.com/HedvigInsurance/market-web/pull/539/files#diff-abf81d5532e69c8f59d134709ec987e19669f348220d8a39cc11e1e722db919bL138-L156)), would say rather you should remove [the field](https://github.com/HedvigInsurance/market-web/blob/207eefe89fda01e74a95011d6c6157f7ac9db250/storyblok/components.json#L2242-L2246) from storyblok.

Hope y'all are ok over there 🤗 the the others i said hi.

robinandeer:
Thanks for the input, @palmenhq! We've still got a lot of "palmen"-accounts to clean up 😅 

I agree that it's reasonable to believe that we could use `noindex` to exclude pages from search AND sitemap.

Will double-check that we are not using that field and look into how to remove it instead.

We are all good here - hope the same goes for you 😌

styrelius:
Yeah thank you @palmenhq ! Also I was like hm who removed that filtering and why and yes it was me and I don't remember why haha 😅 A lot happened in that PR.

Really nice to hear from you, I hope you're doing well!! 🤗

palmenhq:
> Yeah thank you @palmenhq ! Also I was like hm who removed that filtering and why and yes it was me and I don't remember why haha 😅 A lot happened in that PR.

Haha yes, might have been a merge conflict accident also, considering it disappeared in `Merge branch "master"` 😅 But it makes sense i believe

> Really nice to hear from you, I hope you're doing well!! 🤗

Thanks, yes im great thanks :D same to you!

styrelius:
@robinandeer Should we close this one?

robinandeer:
Close in favour of #592

PR #590, opened 2021-07-21
Get image dimensions from URL

robinandeer:
## What?

Spread width/height on image element if it's part of the asset URL.

## Why?

We want to tell the browser the dimensions of the images to avoid layout shifts.

> QUESTION: I can't work out how to integrate this with the `DeferredImage` component. I'm not really sure why it swaps images on the fly and sets the width/height to undefined.

### DIFF
--- a/src/blocks/ImageBlockBrandPivot/ImageBlockBrandPivot.tsx
+++ b/src/blocks/ImageBlockBrandPivot/ImageBlockBrandPivot.tsx
@@ -3,7 +3,11 @@ import React from 'react'
 import { Caption } from 'components/Caption'
 import { ContentWrapper } from '../../components/blockHelpers'
 import { DeferredImage } from '../../components/DeferredImage'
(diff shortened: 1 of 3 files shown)


gustaveen:
I understand that this is to prevent content shifts right? But couldn't see that the widh and height attribute was added to the img tags? (They could get the value from the url in your function). But if that would be added the img ratio would be skewed for responsive images right? Wouldn't you need add an aspect ratio in css then?

robinandeer:
I think it should be combined with something like this:

```
img {
  max-width: 100%;
  height: auto;
}
```

I don't remember if I double-checked that we have added this already though. But in that case I think the images would still resize correctly for all sizes.

robinandeer:
...but the question really is why this happens:

```javascript
// DeferredImage.tsx

<Img
  width={isVisible ? undefined : this.state.width}
  height={isVisible ? undefined : this.state.height}
/>
```

The PR doesn't give us many clues 😅 
https://github.com/HedvigInsurance/market-web/pull/129

PR #592, opened 2021-08-03
Remove "public" Storyblok field

robinandeer:
## What?

Update sitemap test data.
Update Storyblok Story type.

## Why?

This field is no longer used for anything.
Background: https://github.com/HedvigInsurance/market-web/pull/589#issuecomment-891598563


styrelius (approved):
👍 🧹

PR #594, opened 2021-08-09
Generate structured data rich google results

robinandeer:
## What?

Implement FAQPage module for Accordion blocks.
Clean up structured data implementation.
Add SoftwareApplication module.

## Why?

We want to stand out in the search results.

We currently add this from GTM but that's not so manageable so we should do it dynamically here instead.

Reference: https://neilpatel.com/blog/get-started-using-schema/

_Referenced ticket: [TSEO-16]_

[TSEO-16]: https://hedvig.atlassian.net/browse/TSEO-16

## Questions

I'm not 100% if we should include the `SoftwareApplication` module. What do you guys think? Perhaps it only makes sense on the front page.

## Caveats

I noticed that existing structured data is not localised -- we should fix this but I think it would be better to tackle that in a separate PR.

## More

This might be something to look into later: https://github.com/google/react-schemaorg
Try how it works here: https://search.google.com/test/rich-results?utm_campaign=devsite&utm_medium=microdata&utm_source=faq-page

### DIFF
--- a/src/utils/structuredData.ts
+++ b/src/utils/structuredData.ts
@@ -1,0 +1,75 @@
+import { AccordionProps } from 'blocks/AccordionBlock/AccordionBlock'
+
+export const structuredFAQPage = (
+  accordions: ReadonlyArray<AccordionProps>,
+) => ({
+  '@context': 'http://schema.org',
+  '@type': 'FAQPage',
+  mainEntity: accordions.map((item) => ({
+    '@type': 'Question',
+    name: item.title,
+    acceptedAnswer: {
+      '@type': 'Answer',
+      text: item.paragraph.html,
+    },
+  })),
+})
+
+export const structuredSoftwareApplication = () => [
+  {
+    '@context': 'http://schema.org',
+    '@type': 'SoftwareApplication',
+    name: 'Hedvig',
+    operatingSystem: 'IOS',
+    applicationCategory: 'FinanceApplication',
+    aggregateRating: {
+      '@type': 'AggregateRating',
+      ratingValue: '4.7',
+      ratingCount: '1200',
+    },
+  },
+  {
+    '@context': 'http://schema.org',
+    '@type': 'SoftwareApplication',
+    name: 'Hedvig',
+    operatingSystem: 'ANDROID',
+    applicationCategory: 'FinanceApplication',
+    aggregateRating: {
+      '@type': 'AggregateRating',
+      ratingValue: '4.4',
+      ratingCount: '180',
+    },
+  },
+]
+
+export const structuredWebSite = () => ({
+  '@context': 'http://schema.org',
+  '@type': 'WebSite',
+  name: 'Hedvig',
+  url: 'https://www.hedvig.com',
+  description:
+    'Hedvig är en ny typ av försäkring. Byggd på smart teknik, omtanke och sunt förnuft. Så att du kan få hjälp på sekunder, och ersättning på minuter.',
+})
+
+export const structuredOrganization = () => ({
+  '@context': 'http://schema.org',
+  '@type': 'Organization',
+  name: 'Hedvig',
+  url: 'https://www.hedvig.com',
+  logo: 'https://www.hedvig.com/assets-next/favicons/apple-icon.png',
+  description:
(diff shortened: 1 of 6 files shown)


src/utils/structuredData.ts
```diff
@@ -0,0 +1,75 @@
...
+  {
+    '@context': 'http://schema.org',
+    '@type': 'SoftwareApplication',
+    name: 'Hedvig',
+    operatingSystem: 'IOS',
+    applicationCategory: 'FinanceApplication',
+    aggregateRating: {
+      '@type': 'AggregateRating',
+      ratingValue: '4.7',
+      ratingCount: '1200',
```
styrelius, line 28: Is there a way to not hard code these? Feels like something we most definitely will forget to update so it will pretty soon be no longer up-to-date?

src/utils/structuredData.ts
```diff
@@ -0,0 +1,75 @@
...
+  },
+]
+
+export const structuredWebSite = () => ({
+  '@context': 'http://schema.org',
+  '@type': 'WebSite',
+  name: 'Hedvig',
+  url: 'https://www.hedvig.com',
+  description:
+    'Hedvig är en ny typ av försäkring. Byggd på smart teknik, omtanke och sunt förnuft. Så att du kan få hjälp på sekunder, och ersättning på minuter.',
```
styrelius, line 51: Might there be a problem that this description is in Swedish but the `url` is the general for every market? Do we need to have one of these things for each locale?

src/utils/structuredData.ts
robinandeer: Yeah for sure. It's kind of arbitrary that we are allowed to set the values ourselves.

But I'm okey with removing these as well for this PR if we want to investigate it some more!

src/utils/structuredData.ts
robinandeer: Agreed, I'm only replacing what we currently have in this PR so we can merge it quickly :-)

src/utils/structuredData.ts
```diff
@@ -0,0 +1,75 @@
...
+  accordions: ReadonlyArray<AccordionProps>,
+) => ({
+  '@context': 'http://schema.org',
+  '@type': 'FAQPage',
+  mainEntity: accordions.map((item) => ({
+    '@type': 'Question',
+    name: item.title,
+    acceptedAnswer: {
+      '@type': 'Answer',
+      text: item.paragraph.html,
```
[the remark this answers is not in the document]
robinandeer: 1. "make sure that we don't break this" - TypeScript would warn, no? Unless we forget to also update the types but that would break a lot of things on the site.

    I do think we should look into more of a model-view-controller setup where we avoid passing Storyblok data around in the whole code base.

2. "Is it only used for FAQ" - Interesting. I guess the naming of the block suggests that even if I haven't seen it use in a different way...

    Should we perhaps add a checkbox in Storyblok to indicate that it's an FAQ? Or can it be derived some other way?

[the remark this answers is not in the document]
gustaveen: True, we use Accordions on the jobs page https://www.hedvig.com/se-en/jobs :)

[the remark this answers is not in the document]
gustaveen: I think adding a checkbox for adding schema data could work. Since the presentation, an accordion in this case, is the same for all usage. 👍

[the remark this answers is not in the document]
robinandeer: Have added a field in Storyblok for this!

robinandeer:
New field in Storyblok:

Screenshot 2021-08-11 at 14 56 29

robinandeer:
Now you can update structured data description from Storyblok as well!

What do you think of this solution?

Screenshot 2021-08-11 at 18 53 06

gustaveen:
> Now you can update structured data description from Storyblok as well!
> 
> What do you think of this solution with new fields in the "Global Story"?
> 
> Screenshot 2021-08-11 at 18 53 06

Me likey! Perfect to have it as a global setting 👍

gustaveen (approved):
GJ! 🙌

src/utils/structuredData.ts
styrelius: Hm, yeah I would say we could add these later if we can find a way to get those ratings directly from App Store/Google Play

PR #596, opened 2021-08-12
Add review snippet to structured data

robinandeer:
## What?

Add review snippet to structured data for our Trustpilot score:
https://developers.google.com/search/docs/advanced/structured-data/review-snippet

## Why?

Display in Google search results.

## Next step

We can fetch the information from this endpoint:

https://api.trustpilot.com/v1/business-units/5b62ebf41788620001d3c4ae?apikey=XXXXX

However, it requires an API key so we should perform that request from the backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,3 +165,22 @@ a) Running 2 commands isn't too bad right? b) 3 real reasons:
    (<- not a biggie but still).
 
 ☝️ Are none of these true anymore? Feel free to automate it!
+
+### Updating Trustpilot score in Storyblok
+
+There's a script that updates the latest Trustpilot score for one market/locale at a time.
+This information is used for the review snippet module that shows up in rich Google search results.
+
+```sh
+yarn update-trustpilot-score --space 123714 --page 67792582
+```
+
+**Requirements**
+
+- `STORYBLOK_MANAGEMENT_TOKEN` Define your Storyblok management token in `.env`
+- `TRUSTPILOT_API_KEY` Define the Trustpilot API key in `.env`
+
+**Flags**
+
+- `--space N` Space to fetch.
+- `--page N` Page ID for the "global" story in the space.
--- a/README.md
+++ b/README.md
@@ -168,11 +168,10 @@ a) Running 2 commands isn't too bad right? b) 3 real reasons:
 
 ### Updating Trustpilot score in Storyblok
 
-There's a script that updates the latest Trustpilot score for one market/locale at a time.
-This information is used for the review snippet module that shows up in rich Google search results.
+There's a script that updates the latest Trustpilot score for one market/locale at a time. The information is stored in the "Global" component for each locale. The score is used for the review snippet module that shows up in rich Google search results.
 
 ```sh
-yarn update-trustpilot-score --space 123714 --page 67792582
+yarn update-trustpilot-score --space 123714 --page 67792582 --businessunit 5b62ebf41788620001d3c4ae
 ```
 
 **Requirements**
@@ -182,5 +181,6 @@ yarn update-trustpilot-score --space 123714 --page 67792582
 
 **Flags**
 
-- `--space N` Space to fetch.
-- `--page N` Page ID for the "global" story in the space.
+- `--space <id>` Storyblok Space ID to fetch and update.
+- `--page <id>` Storyblok Page ID for the locale-specific "global" story.
+- `--businessunit <id>` [Trustpilot Business Unit ID for Hedvig](https://documentation-apidocumentation.trustpilot.com/#BusinessUnitID)
--- a/src/utils/structuredData.ts
+++ b/src/utils/structuredData.ts
@@ -66,10 +66,9 @@ interface StructuredOrganizationParams {
   description?: string
 }
 
-export const structuredOrganization = ({
-  description = DEFAULT_ORG_DESCRIPTION,
-}: StructuredOrganizationParams = {}) => ({
-  '@context': 'http://schema.org',
+const structuredDataOrganization = ({
+  description,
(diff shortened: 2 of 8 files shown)


styrelius (commented):
If we're not going to use the TrustPilot API to get the current score since that's somewhat a bigger project that involves back-end and so on, I would suggest that the `ratingValue`, the `bestRating` and the `reviewCount` should be added/modified from Storyblok, like we do with the organization description. 
If we have to change it in the actual codebase we most certainly will forget to and the numbers will very soon be very wrong 😬

src/utils/structuredData.ts
```diff
@@ -89,3 +88,21 @@ export const structuredOrganization = ({
...
+  ...structuredDataOrganization({ description }),
+})
+
+export const structuredDataReviewSnippet = ({
+  description = DEFAULT_ORG_DESCRIPTION,
+}: StructuredOrganizationParams = {}) => ({
+  '@context': 'http://schema.org',
+  '@type': 'AggregateRating',
+  itemReviewed: structuredDataOrganization({ description }),
+  ratingValue: '4.8',
```
gustaveen, line 105: Since we have a hardcoded value of our Trustpilot score on the site today I think it's fine to have it here as well for the time being :)

gustaveen (commented):
Yes, since we have a hardcoded value of our Trustpilot score on the site today I think it's fine to have it here as well for the time being or as you suggested added to Storyblok 👍 Best rating will always be 5 I assume so that can be omited probably

robinandeer:
Right, it's probably better to store this information in Storyblok. I can look into that.

robinandeer:
@gustaveen @styrelius I've updated the implementation now.

I investigated directly calling the Trustpilot API to get the values initially. However, the values don't need to be updated very often so I believe it's necessary.

My solution caches the values in Storyblok like we talked about ⬆️ 

I've also added a script that calls the Trustpilot API and updates Storyblok automatically.

Next step could potentially be to run that script on a schedule or e.g. when we deploy the site.

styrelius (commented):
Very cool! 🙌 🙌  I think we need some more documentation on how to run this though. I've been struggling for a good amount of time now with different ID's from Storyblok and TrustPilot and have still not been able to run the script successfully. 😔

gustaveen (commented):
Great update! 🎉 Works great! Just a question really about the business unit ID, is it unique for all markets or is it the same? 
Minor suggesion, maybe group the Trustpilot settings under a Trustpilot folder? 
Screenshot 2021-08-24 at 14 19 20

robinandeer:
@gustaveen 
> Could there be a way to get the Global stories dynamically instead of specifying page ids? :) But that might be overkill.

Yeah there probably is but I think we need to integrate with the delivery API then as well. I think if we integrate with CI and run it automatically, this solution might be easier to maintain. Does that make sense?

robinandeer:
@gustaveen 
> Great update! 🎉 Works great! Just a question really about the business unit ID, is it unique for all markets or is it the same?
> Minor suggesion, maybe group the Trustpilot settings under a Trustpilot folder?

I followed @styrelius suggestion and moved the business unit param to the command line. It's not something editors should need to worry about so I don't think it belongs in Storyblok on second thought.

Perhaps we don't have different business units for each market. However, do we have any place to store such cross-market global values?

README.md
```diff
@@ -165,3 +165,22 @@ a) Running 2 commands isn't too bad right? b) 3 real reasons:
...
+
+**Requirements**
+
+- `STORYBLOK_MANAGEMENT_TOKEN` Define your Storyblok management token in `.env`
+- `TRUSTPILOT_API_KEY` Define the Trustpilot API key in `.env`
+
+**Flags**
+
+- `--space <id>` Storyblok Space ID to fetch and update.
+- `--page <id>` Storyblok Page ID for the locale-specific "global" story.
```
styrelius, line 185: 😍

styrelius:
I agree with @gustaveen that it would be neat to group the Trustpilot settings under a Trustpilot folder in Storyblok - to make it clearer what it is for the editors of Storyblok. Also maybe add a comment that they shouldn't touch these fields since they are updated from the TrustPilot API?

README.md
```diff
@@ -168,11 +168,10 @@ a) Running 2 commands isn't too bad right? b) 3 real reasons:
 
 ### Updating Trustpilot score in Storyblok
 
-There's a script that updates the latest Trustpilot score for one market/locale at a time.
-This information is used for the review snippet module that shows up in rich Google search results.
+There's a script that updates the latest Trustpilot score for one market/locale at a time. The information is stored in the "Global" component for each locale. The score is used for the review snippet module that shows up in rich Google search results.
 
 ```sh
-yarn update-trustpilot-score --space 123714 --page 67792582
+yarn update-trustpilot-score --space 123714 --page 67792582 --businessunit 5b62ebf41788620001d3c4ae
```
gustaveen, line 174: Great! Makes perfect sense 🙌

gustaveen (commented):
> However, do we have any place to store such cross-market global values?

No we don't, so that's what I realised might be nice to have longterm, if there will be a bunch of repeated settings on all global setting pages going forward :)

gustaveen (dismissed):
🙌

styrelius (commented):
Nice! Still think that it's a good idea like Gustav wrote to group these review things under a **Trustpilot** folder, like we do with e.g. **Structured Data** (i.e. add a `section` to `storyblok/components.json`), and also have a little description for each of these input fields so that the editors of Storyblok know that they shouldn't manually update these fields. By the way could they be "locked"/disabled in some way so can't edit them manually?

Also it seems like these fields are lacking the `pos` property? 

image

gustaveen (approved):
⭐️

PR #598, opened 2021-08-31
First stab at removing constate

gustaveen:
## What?
First stab at removing constate and use regular react context. Adding a context for providing `currentLocale`.

## Why?
I need to access market to send to GTM. So instead of hacking away at constate, let's get rid off it 🙏

This PR doesn't fix the provider for storyblok data.


src/context/LocaleContext/useLocal.ts
```diff
@@ -0,0 +1,10 @@
+import { useContext } from 'react'
+import { LocaleContext } from './LocalContext'
+
+export const useLocal = () => {
```
styrelius, line 4: Shouldn't this, and the file, called `useLocale` (missing an "e" at the end)?

Also, I think this could probably be in the `LocaleContext.ts` file since it's so tightly connected to that and makes no sense on its own? 🤔

src/context/LocaleContext/LocalContext.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { createContext } from 'react'
```
styrelius, line 1: I assume the file should be called `LocaleContext.tsx` (Local**e** instead of Local), and that this is just a typo?

styrelius (commented):
Really nice! 🙌 😀 

Interestingly enough Sebastian Markbåge on the React core team tweeted something a while back about that this is not the way one should use React context (although this is exactly how everyone uses it lol), but that a context should always work meaningfully with its default value(s). That being said I still think we should do it like this since there isn't really a simple alternative and these are old tweets but just wanting to drop that confusing fact 🙃 

[https://twitter.com/sebmarkbage/status/1219828998282215430?s=24](https://twitter.com/sebmarkbage/status/1219828998282215430?s=24)
[https://twitter.com/sebmarkbage/status/1219836431972978689?s=24](https://twitter.com/sebmarkbage/status/1219836431972978689?s=24)

src/context/LocaleContext/useLocal.ts
gustaveen: Haha, that is a typo, thanks! 😂 🙌 
I'm just used to writing hooks in it's own file I think, but I'm good with either solution 👍

gustaveen:
> Really nice! 🙌 😀
> 
> Interestingly enough Sebastian Markbåge on the React core team tweeted something a while back about that this is not the way one should use React context (although this is exactly how everyone uses it lol), but that a context should always work meaningfully with its default value(s). That being said I still think we should do it like this since there isn't really a simple alternative and these are old tweets but just wanting to drop that confusing fact 🙃
> 
> https://twitter.com/sebmarkbage/status/1219828998282215430?s=24
> https://twitter.com/sebmarkbage/status/1219836431972978689?s=24

I hear you! I was a bit unsure what to set the default value to but felt it didn't make sense to set it to s specific locale. I mean in the example he talks about Dark mode/ Light mode it makes perfect sense to have a set defaultValue. But in other cases not so much. So I did what Kenta taught me 😄 https://kentcdodds.com/blog/how-to-use-react-context-effectively#typescript

gustaveen:
And we can for sure look into if we don't want to use context for these type of things. Felt like the first step to get rid of constate though :)

robinandeer:
Great that you are looking into cleaning this up :)

I also heard about those tweets from a podcast episode I recommended to @gustaveen earlier: https://frontendfirst.fm/episodes/shared-reactive-data-without-context-or-effects

I think this sounds like one of those cases where context (or any React state) isn't necessary. The locale-value is just set once and never updated for the life time of the app. It could live in module scope!

But I agree that we can see this as step one. I would not be against setting a default locale though, e.g. `en-se` to make the site work meaningfully with a default context.

styrelius:
Ah, I also heard about it on Frontend First 😄 That blog post on context from Kenta is great though, and I think we should keep this. For now at least.

src/context/LocaleContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export const LocaleContext = createContext<LocaleContext>({
+  currentLocale: locales['se-en'],
+})
+
+export const useLocale = () => {
+  const context = useContext(LocaleContext)
+  if (context === undefined) {
+    throw new Error('useLocale has to be used inside a LocaleProvider')
+  }
```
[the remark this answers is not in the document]
gustaveen: Ah I thought you meant that I should set the default so we could remove it from the test. :) Because the test didn't fail now when we have a `defaultvalue` funny enough. But let's put it back. And sure, we can remove the check 🙌

robinandeer (approved):
Looks good!

On second thought, I think it's better the test setup is mimicking the actual setup when it comes to global providers so this feels better.

PR #599, opened 2021-09-02
Feature/tracking/user properties

gustaveen:
## What?
Track user properties

Branched of https://github.com/HedvigInsurance/market-web/pull/598

## Why?
We need to track the same user properties in marker-web as we do in web-onboarding

Screenshot 2021-09-02 at 10 50 38

### DIFF
--- a/src/context/LocaleContext.tsx
+++ b/src/context/LocaleContext.tsx
@@ -1,17 +1,17 @@
 import React, { createContext, useContext } from 'react'
 import { LocaleData, locales } from 'utils/locales'
 
-export type LocaleContext = {
+type LocaleContextProps = {
   currentLocale: LocaleData
 }
 
-export const LocaleContext = createContext<LocaleContext>({
+const LocaleContext = createContext<LocaleContextProps>({
   currentLocale: locales['se-en'],
 })
 
 export const useLocale = () => useContext(LocaleContext)
 
-export const LocaleProvider: React.FC<LocaleContext> = ({
+export const LocaleProvider: React.FC<LocaleContextProps> = ({
   currentLocale,
   children,
 }) => {
--- a/src/context/LocaleContext.tsx
+++ b/src/context/LocaleContext.tsx
@@ -9,13 +9,7 @@ export const LocaleContext = createContext<LocaleContext>({
   currentLocale: locales['se-en'],
 })
 
-export const useLocale = () => {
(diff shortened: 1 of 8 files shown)


src/utils/tracking/gtm.ts
```diff
@@ -0,0 +1,48 @@
+import { useEffect } from 'react'
+import { useLocale } from 'context/LocaleContext'
+export type AppEnvironment = 'development' | 'staging' | 'production'
+
+type GTMUserProperties = {
+  market: string
+  environment: AppEnvironment | undefined
```
robinandeer, line 7:
```suggestion
  environment?: AppEnvironment
```

src/server/utils/tracking.ts
```diff
@@ -17,6 +17,10 @@ const snap = `
 `
 
 export const allTracking = (nonce: string) => `
+<script nonce="${nonce}">
+  dataLayer = [];
```
robinandeer, line 21: Is it more clear to write it out like:

```suggestion
  window.dataLayer = [];
```

robinandeer (dismissed):
It's slightly confusing since you have the `useLocale` changes in here also 😅 but I think it looks good!

gustaveen:
> It's slightly confusing since you have the `useLocale` changes in here also 😅 but I think it looks good!

Ah sorry for that :D Will merge the other PR first 🙃

src/server/utils/tracking.ts
gustaveen: I kinda like to stick to the dev docs from google https://developers.google.com/tag-manager/devguide#adding-data-layer-variables-to-a-page :)

src/context/LocaleContext.tsx
```diff
@@ -9,13 +9,7 @@ export const LocaleContext = createContext<LocaleContext>({
   currentLocale: locales['se-en'],
```
styrelius, line 9: Sorry this is out of context for this PR but does the `LocaleContext` need to be exported?

styrelius (commented):
Looks very good to me! Just one thing that's actually unrelated to this 🙈

src/context/LocaleContext.tsx
gustaveen: Oh, not at all, well spotted! 🙏

styrelius (approved):
🛤️ 👑 ✨

PR #606, opened 2021-10-06
Update node to latest LTS (Long term support)

gustaveen:
## What?
It's actually at 14.18.0 right now but types and nvm is lagging behind

## Why?
Keep us up to date :)


.ci/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14.16.0
+FROM node:14.17.6
```
robinandeer, line 1: Is this used by Heroku?

robinandeer (approved):
Nice!

Do we know how/if we need to update CodeFresh and Heroku to use the new Node.js version?

PR #607, opened 2021-10-07
Update perils query and add Accident as an option

gustaveen:
## What?

- Update query from `perils` to `contractPerils`. The new url is absolute and will go to a bucket on S3.
- Add Accident contract types to Perils block

@joacimastrom is updating the CORS permissions for icons, since they were initially blocked. 

## Why?
Perils are no longer fetched from content-service but should be fetched from the new Promise CMS

We're currently live with these pages so would be good to get out asap :) 

Screenshot 2021-10-07 at 16 04 06

### DIFF
--- a/storyblok/components.json
+++ b/storyblok/components.json
@@ -2193,6 +2193,10 @@
               "name": "SE_APARTMENT_STUDENT_RENT"
             },
             {
+              "value": "SE_ACCIDENT",
+              "name": "SE_ACCIDENT"
+            },
+            {
               "value": "NO_HOME_CONTENT_OWN",
               "name": "NO_HOME_CONTENT_OWN"
             },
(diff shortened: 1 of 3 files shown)


storyblok/components.json
```diff
@@ -2192,6 +2192,14 @@
...
               "name": "SE_APARTMENT_STUDENT_RENT"
             },
+            {
+              "value": "SE_ACCIDENT",
+              "name": "SE_ACCIDENT"
+            },
+            {
+              "value": "SE_ACCIDENT_STUDENT",
+              "name": "SE_ACCIDENT_STUDENT"
+            },
```
styrelius, line 2202: Do we need two different options here? I was under the impression that the student accident insurance is exactly the same as the regular one? But I might have misunderstood of course 😅

styrelius (commented):
Awesome, just one question 👇

mehdielket (dismissed):
Seems minimalistic and all good to me. Let me know if you need help to test on staging :)

gustaveen:
Thanks guys! Just removed the student option since we won't use it on Market web I guess :) Feel free to merge!

styrelius (approved):
👯‍♂️

PR #608, opened 2021-10-18
Add support for fourth & fifth link in InsuranceInfoBlock

gustaveen:
## What?
Add support for fourth and fifth link in InsuranceInfoBlock

## Why?
We need to display more T&Cs for compliance reasons

This should really be a repeatable field instead. But that would mean a few of content updates, will check with Mange if we could coordinate that so we could refactor this

Screenshot 2021-10-18 at 10 29 39


robinandeer (approved):
Nice 🎉 

una mas

PR #610, opened 2021-11-15
Improve locales module

styrelius:
## What?

1) `locales.ts` is moved to a new folder `/l10n` directly under `/src`
2) Stricter typing of `locales` object, with union types for all possible values
3) Stuff that's not the `locales` object or its types (like helper/utils functions) is moved to its own file: `/l10n/l10n-utils.ts`

## Why?

1) Made more sense to me to have `locales.ts` located in a dedicated folder instead of in the `/src/utils` folder 
2) The idea behind those new union types is that even though it might be annoying to add your values twice when adding a new locale (since you need to add it both to the respective union types _and_ to the actual `locales` object) it makes the risk of typos much lower.
3) With mire markets added `locales.ts` is at a risk of becoming messy :)


src/l10n/locales.ts
```diff
@@ -1,10 +1,16 @@
...
+type MarketName =
+  | 'Sverige'
+  | 'Sweden'
+  | 'Norge'
+  | 'Norway'
+  | 'Danmark'
+  | 'Denmark'
+type LanguageLabel = 'Sv' | 'En' | 'No' | 'Da'
+type HtmlLang = 'sv' | 'en' | 'no' | 'da'
+type HrefLang = 'sv-se' | 'en-se' | 'no-no' | 'en-no' | 'da-dk' | 'en-dk'
```
robinandeer, line 13: 🎉

robinandeer (approved):
It's just a very nice improvement

PR #611, opened 2021-11-15
GRW-524 - Freature: added locale support for France 🇫🇷 

guilhermespopolin:
## What?

- `locales.ts` file was updated in order to support launch in France

## Why?

- Because of reasons 😝 

**Referenced ticket(s): [GRW-524](https://hedvig.atlassian.net/browse/GRW-524)**

### DIFF
--- a/src/l10n/locales.ts
+++ b/src/l10n/locales.ts
@@ -16,8 +16,7 @@ type MarketName =
   | 'Norway'
   | 'Danmark'
   | 'Denmark'
-  | 'Français'
-  | 'French'
+  | 'France'
 type LanguageLabel = 'Sv' | 'En' | 'No' | 'Da' | 'Fr'
 type HtmlLang = 'sv' | 'en' | 'no' | 'da' | 'fr'
 type HrefLang =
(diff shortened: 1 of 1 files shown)


src/l10n/locales.ts
```diff
@@ -93,4 +111,22 @@ export const locales: Locales = {
...
+    marketName: 'Français',
+    langLabel: 'Fr',
+    htmlLang: 'fr',
+    hrefLang: 'fr-fr',
+  },
+  'fr-en': {
+    label: 'fr-en',
+    iso: 'en-FR',
+    marketLabel: 'fr',
+    marketName: 'French',
```
gustaveen, line 127:
```suggestion
    marketName: 'France',
```

gustaveen (commented):
🙌 🇫🇷

src/l10n/locales.ts
```diff
@@ -93,4 +111,22 @@ export const locales: Locales = {
     adtractionSrc: 'https://cdn.adt387.com/jsTag?ap=1589794294',
     trustpilotLocale: 'en-US',
   },
+  fr: {
+    label: 'fr',
+    iso: 'fr-FR',
+    marketLabel: 'fr',
+    marketName: 'Français',
```
gustaveen, line 118: This is rather the name of the market and not the language :) Don't remember my school fre that good if it should even be _La France_ but I guess France is sufficient 🤓
```suggestion
    marketName: 'France',
```

gustaveen (approved):
🇫🇷

PR #614, opened 2021-11-19
update twitter meta-image

dendoesit:
## What?

Changed the way the meta-image for twitter is server

## Why?

Twitter card validator cannot render an image. 
My guess is because currently it comes from //a.storyblok and I believe twitter is not seeing the url as it should properly be with https://hedvig.com/******

[**Referenced ticket(s): []**](https://hedvig.atlassian.net/browse/GRW-593)

image


gustaveen (approved):
Aha, nice find Dan! 🙌

PR #616, opened 2021-11-30
Remove V2 colors

gustaveen:
## What?
Remove deprecated colors from brand

## Why?
So we can clean up brand repo


robinandeer (approved):
Nice stuff!

I more and more like this object syntax 😅

PR #618, opened 2021-12-14
Add duplicated block to enable renaming and migration

gustaveen:
## What?
Add duplicate of `BulletPointBlockBrandPivot` without BrandPivot suffix. We need to keep the old one in the transition between component usage. Tis enable us to rename the component from `bullet_point_block_brand_pivot` to `bullet_point_block` without the need to migrate content.

## Why?
To remove BrandPivot suffix of the component. In order to do that we need to rename the component in the the space component dashboard. Storyblok will automatically migrate to the new name (see screenshot). We can't update `components.json` before the migration of content is done in the actual space. Works fine locally and will double check in staging. When doing this in prod I will also lock the editing capabilities for ediotors.

Screenshot 2021-12-14 at 14 48 04

Didn't know this had been a feature for a long time :) https://www.storyblok.com/cl/component-renamer

### DIFF
--- a/src/blocks/BulletPointBlock.tsx
+++ b/src/blocks/BulletPointBlock.tsx
@@ -1,0 +1,187 @@
+import styled from '@emotion/styled'
+import React from 'react'
+import {
+  ContentWrapper,
+  getMinimalColorStyles,
+  LAPTOP_BP_UP,
+  MOBILE_BP_DOWN,
+  MOBILE_BP_UP,
+  SectionWrapper,
+  TABLET_BP_UP,
+} from 'components/blockHelpers'
+import { DeferredImage } from 'components/DeferredImage'
+import { getStoryblokImage, Image } from 'utils/storyblok'
+import {
+  BaseBlockProps,
+  MarkdownHtmlComponent,
+  MinimalColorComponent,
+} from 'blocks/BaseBlockProps'
+
+const BulletPointSectionWrapper = styled(SectionWrapper)({
+  overflowX: 'hidden',
+})
+
+const InnerWrapper = styled('div')({
+  display: 'flex',
+  flexWrap: 'wrap',
+  minWidth: '100%',
+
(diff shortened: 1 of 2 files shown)


src/blocks/BulletPointBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+} from 'blocks/BaseBlockProps'
+
+const BulletPointSectionWrapper = styled(SectionWrapper)({
+  overflowX: 'hidden',
+})
+
+const InnerWrapper = styled('div')({
+  display: 'flex',
+  flexWrap: 'wrap',
+  minWidth: '100%',
```
robinandeer, line 27: Maybe a task for @styrelius to dig into if we should use object syntax or template literals 😉

src/blocks/BulletPointBlock.tsx
```diff
@@ -0,0 +1,187 @@
...
+  },
+})
+
+const BulletPointBody = styled('div')<{
+  alignCenter: boolean
+  colorComponent: MinimalColorComponent
+}>(({ alignCenter, colorComponent }) => ({
+  maxWidth: alignCenter ? '16rem' : 'none',
+  color: getMinimalColorStyles(colorComponent?.color ?? 'standard').color,
+  fontSize: alignCenter ? '18px' : '16px',
```
robinandeer, line 120: nit: why no rems?

robinandeer (approved):
Nice code scouting

src/blocks/BulletPointBlock.tsx
gustaveen: I just did a full copy of the component so didn't change anything. I will go about it when I start to update the component in the nest PR :) @robinandeer

src/blocks/BulletPointBlock.tsx
gustaveen: hehe, I have slight favour towards one of them ;)

PR #621, opened 2021-12-22
Update Bullet Point Block

gustaveen:
## What?
Update BulletPointBlock with new features:

- [x] Add component to Storybook
- [x] Add layout support for images
- [x] Add support for dynamic number of columns
- [x] Add support for multiple heading sizes

Also added Storybook assets like icons and images to use as examples in stories 

## Why?
There is a need for more flexibility 

https://user-images.githubusercontent.com/6661511/147116810-e61dfc10-17d9-409b-895e-1a312adae255.mov

https://www.figma.com/file/W3f1Ff6zcsAG6LP3PMKvtO/StoryBlok-fixes-before-DK-launch-%3A)?node-id=418%3A1749


src/blocks/BulletPointBlock/BulletPointBlock.tsx
```diff
@@ -0,0 +1,217 @@
...
+>
+
+export type BulletPointBlockProps = BrandPivotBaseBlockProps & {
+  align_center: boolean
+  bullet_point_layout: boolean
+  color_body: MinimalColorComponent
+  bullet_points: BulletPointItemProps
+}
+
+export const BulletPointBlock: React.FC<BulletPointBlockProps> = ({
```
styrelius, line 157: I think we said we're moving away from using return types (`React.FC`) for components, since that's apparently [discouraged](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components)?

src/blocks/BulletPointBlock/BulletPointBlock.tsx
```diff
@@ -0,0 +1,217 @@
...
+          max-width: none;
+          max-height: none;
+          width: auto;
+          margin-right: 0;
+        }
+      `}
+`
+
+const BulletPointImage = styled(DeferredImage)<{
+  alignCenter: boolean
```
styrelius, line 113: This prop doesn't seem to be used?

src/blocks/BulletPointBlock/BulletPointBlock.tsx
```diff
@@ -0,0 +1,217 @@
...
+  flex-wrap: wrap;
+  min-width: 100%;
+
+  ${MOBILE_BP_DOWN} {
+    padding-left: 8px;
+    padding-right: 8px;
+  }
+
+  ${MOBILE_BP_UP} {
+    margin-left: -1.5rem;
```
styrelius, line 36: Why do we need the minus margins? Maybe it's just me but I always get a little nervous by minus margins, I feel like they can mess things up 😅

src/blocks/BulletPointBlock/BulletPointBlock.tsx
```diff
@@ -0,0 +1,217 @@
...
+  margin-bottom: 1rem;
+`
+
+const BulletPoint = styled.div<{
+  alignCenter: boolean
+  bulletPointLayout: boolean
+  columns: number
+}>`
+  display: flex;
+  flex-direction: ${(props) => (props.bulletPointLayout ? 'row' : 'column')};
```
styrelius, line 54: [Nit/personal preference]: Maybe destructure the props that are used in the `styled` components? To me that's more readable :)
```suggestion
  flex-direction: ${({ bulletPointLayout }) => (bulletPointLayout ? 'row' : 'column')};
```

src/blocks/BulletPointBlock/BulletPointBlock.tsx
```diff
@@ -0,0 +1,217 @@
...
+      `}
+  }
+
+  ${LAPTOP_BP_UP} {
+    width: ${(props) => `calc((1 / ${props.columns} * 100%) - 3.5rem)`};
+    margin-left: 3.5rem;
+  }
+`
+
+const BulletPointHead = styled.div<{
```
styrelius, line 86: This name isn't super clear to me 🤔 Is it only used to wrap the image? Then maybe the name could be something that explains that? Or what's the "Head"? 🙈

src/blocks/BulletPointBlock/BulletPointBlock.tsx
```diff
@@ -0,0 +1,217 @@
...
+}) => (
+  <BulletPointSectionWrapper
+    brandPivot
+    colorComponent={color}
+    size={size}
+    extraStyling={extra_styling}
+  >
+    <ContentWrapper fullWidth>
+      <InnerWrapper>
+        {bullet_points.map((bullet) => (
```
styrelius, line 174: Maybe destructure? 😀 (Yes I'm obsessed with destructuring)
```suggestion
        {bullet_points.map(({ uuid, image, title, title_size, title_size_mobile, paragraph }) => (
```

styrelius (commented):
Great job with this block!! 👑

src/blocks/BulletPointBlock/BulletPointBlock.tsx
gustaveen: Cool, didn't know that, always nice to get rid of boilerplate 🙌 When did we decide that? Would be nice if there was a lint rule for it. There seem to be a codemod for it so could be nice to fix for all occurrences https://github.com/gndelia/codemod-replace-react-fc-typescript 
Maybe something for maintainance day :)

src/blocks/BulletPointBlock/BulletPointBlock.tsx
gustaveen: Ah, I moved the existing component so most of the component is old code. This is a solution we use to make the column layout with `flexbox` to get tha gap between items. Each item has a `margin-left` and then you offset the wrapper with the same negative value to center it.

src/blocks/BulletPointBlock/BulletPointBlock.tsx
gustaveen: I agree, `BulletPointHeader` would make more sense IMO :)

src/blocks/BulletPointBlock/BulletPointBlock.tsx
gustaveen: I think I'm so used to writing `props` that I find it easier to read :D

src/blocks/BulletPointBlock/BulletPointBlock.tsx
gustaveen: I agree that destructuring is the way to go most times, but when it's inside a component like I prefer to not do it. Because then I think it's easier to see where props comes from and it's a bit more condensed and easier to read :)

src/blocks/BulletPointBlock/BulletPointBlock.tsx
styrelius: Ah OK, I see! Well I'm probably just too nervous generally 😅

src/blocks/BulletPointBlock/BulletPointBlock.tsx
styrelius: I think that was a decision made in some PR discussion so probably not the best way to make decisions lol 🙈 

Great ideas, both adding linting rule and doing the codemod replacement on maintenance day 👌

src/blocks/BulletPointBlock/BulletPointBlock.tsx
styrelius: Haha OK, we obviously have very different preferences for readability 🙃

src/blocks/BulletPointBlock/BulletPointBlock.tsx
styrelius: Oh, OK. It's just so much more... characters that it gets blurry to me. 

Anyway, this might be something for the code style guide, since for example my impulse is always to refactor to destructuring syntax (did it just a few days ago in the `Button` component in `web-onboarding`) and it's probably a waste of time if we'd keep refactoring back and forth haha 😄

src/blocks/BulletPointBlock/BulletPointBlock.stories.tsx
```diff
@@ -0,0 +1,152 @@
+import { select, withKnobs, boolean } from '@storybook/addon-knobs'
```
styrelius, line 1: OK so this is nothing we should change now (but maybe add a ticket or plan for maintenance day?), but the `@storybook/addon-knobs` packages is apparently [deprecated](https://www.npmjs.com/package/@storybook/addon-knobs) so we should probably replace that.

styrelius (approved):
Great job! 🥳 🎈✨ 

(Although too little destructuring for my taste)

guilhermespopolin (approved):
Looks great

src/blocks/BulletPointBlock/BulletPointBlock.tsx
gustaveen: Great that you asked! 👌

src/blocks/BulletPointBlock/BulletPointBlock.tsx
gustaveen: Agreed! I could happily use that pattern, so as you say, we could commit to one or the other 🙌

src/blocks/BulletPointBlock/BulletPointBlock.stories.tsx
gustaveen: Yes indeed! @styrelius I tried to update storybook two times but always ran in to some babel error 😅 So maybe good to give it another stab tomorrow on maintenance day or just fix it in the new setup 👍

src/blocks/BulletPointBlock/BulletPointBlock.tsx
```diff
@@ -0,0 +1,212 @@
...
+  color,
+  color_body,
+  size,
+  bullet_points,
+}: BulletPointBlockProps) => (
+  <BulletPointSectionWrapper
+    brandPivot
+    colorComponent={color}
+    size={size}
+    extraStyling={extra_styling}
```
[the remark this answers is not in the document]
gustaveen: Ah yeah, that's the goal for sure! 🎉 The last bit I want from them is to agree on a new max-widths, since that' mostly what they use it for @robinandeer

PR #622, opened 2022-01-05
Update browserlist config

gustaveen:
## What?

- Update browsers data from `caniuse-lite` (more info here: https://github.com/browserslist/browserslist#browsers-data-updating)
- Use a `.browserslistrc` file for configuration. Babel will use that config instead and so will other tools using `browserlist`

I validated the support against our user base: https://datastudio.google.com/reporting/035efb07-6771-4fac-b34c-128a7ed955e3

## Why?
So that the browser data is up to date with todays versions. Using the `defaults` setting also takes us a step towards using less custom setups. The `defaults` give us a good baseline and we can then remove `IE` from that.


robinandeer (approved):
Nice stuff!

guilhermespopolin (approved):

gustaveen:
Sorry, the reason for not merging this was that this error seemed to pop up after I updated the browser config 😄 Can't understand why it would complain over that 
Screenshot 2022-01-05 at 17 42 56

robinandeer (commented):
@gustaveen should we still keep this one?

gustaveen:
> @gustaveen should we still keep this one?

Nah let's close this for now :)

PR #624, opened 2022-01-13
Add seperate GTM scripts for staging/dev and prod

gustaveen:
## What?
Add separate GTM scripts for staging/dev and prod.

## Why?
This enables us to deploy and test tracking in staging a lot easier. We can deploy container versions to test in staging/dev before publishing to prod. 

More info here: https://support.google.com/tagmanager/answer/6311518?hl=en

Screenshot 2022-01-13 at 13 03 45


guilhermespopolin (approved):
It seems good, but to be honest I don't understand much of this area 😬

gustaveen:
@guilhermespopolin hehe I totally get that, this is very detailed google tag manager stuff 😄 but the bottom line is that Sonny and us can test tracking in staging before we do "deploy" in GTM. Before we could use something called a preview mode but that has is limitations and is a bit harder tow work with.

PR #625, opened 2022-01-17
Delete cookie consent component

gustaveen:
## What?
Delete cookie consent component we have today in favour for the new OneTrust cookie banner that we will implement via GTM. Will use this review app for a while to show stakeholders.

## Why?
We will no longer use this cookie consent strategy


robinandeer (approved):
Easy approve hehe (for when you're ready)

guilhermespopolin (approved):
🧹

PR #634, opened 2022-02-14
Delete deprecated image blocks

gustaveen:
## What?
Delete deprecated image blocks

## Why?
More info here https://github.com/HedvigInsurance/market-web/pull/631


storyblok/components.json
```diff
@@ -78,7 +78,9 @@
       "all_presets": [],
       "preset_id": null,
       "real_name": "accordion_block_brand_pivot",
-      "component_group_uuid": null
+      "component_group_uuid": null,
+      "color": null,
+      "icon": null
```
gustaveen, line 83: These are new fields from Storyblok where you can add some options to the admin view

robinandeer (approved):

PR #641, opened 2022-03-01
Test new HedvigLetters font

gustaveen:
## What?
We are getting new, custom made typefaces 💅 HedvigLetters-Standard will replace Favorit as our sans-serif font.
We also have HedvigLetters-Big and HedvigLetters-Small that might be added to web as well for some large headings.

Screenshot 2022-02-28 at 14 29 32

## Why?
This PR is just to try it out in a review app for design proofing from Kim and brand team so we can see how much tweaking that would be needed.

Differences between Favorit and HedvigLetters-Standard. By the looks of it, not much difference in terms of size, leading and kerning. So will probably not need to tweak font rules that much.
https://user-images.githubusercontent.com/6661511/156169517-b79973d8-f78b-4d27-93cd-a5f63146ea5f.mov


src/components/GlobalStyles.tsx
```diff
@@ -1,15 +1,41 @@
 import { css } from '@emotion/core'
-import { colorsV3, fonts, getCdnFontFaces } from '@hedviginsurance/brand'
+import { colorsV3, getCdnFontFaces } from '@hedviginsurance/brand'
 
 export const globalStyles = css`
   ${getCdnFontFaces()}
```
robinandeer, line 5: Do we need these as well?

robinandeer (approved):
Nice :D

guilhermespopolin (approved):

src/components/GlobalStyles.tsx
gustaveen: Ah yes, when we get the from the brand package for sure 🙌

gustaveen:
Thanks @robinandeer and @guilhermespopolin 🙌 I will probably not merge this, since I will add it to the brand repo. So if you can have a look at this PR that would be awesome 🤗

PR #644, opened 2022-03-09
GRW-792 Add heading tag option

gustaveen:
## What?
Possibility to pick heading tag for `ImageTextBlock `

## Why?
When the block contains the main title of the page, it needs to be an h1

Screenshot 2022-03-09 at 14 44 41

### DIFF
--- a/storyblok/components.json
+++ b/storyblok/components.json
@@ -1700,6 +1700,7 @@
             "title_size",
             "title_size_mobile",
             "title_color",
+            "heading_tag",
             "show_hedvig_wordmark",
             "paragraph",
             "text_position",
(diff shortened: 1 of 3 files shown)


storyblok/components.json
```diff
@@ -1778,18 +1779,43 @@
           "options": [],
           "pos": 4
         },
+        "heading_tag": {
```
guilhermespopolin, line 1782: Just out of curiosity is this file auto generated?

guilhermespopolin (approved):

storyblok/components.json
gustaveen: Good question! We have a script to fetch components updates (when you add a field in your own demo space in Storyblok CMS during local development). 
### Fetch space

If you update components in Storyblok, this is your way to get them into the repository. Will overwrite
`storyblok/components.json`. Make sure you do this if you want to get a schema change into production or staging!

```bash
bin/fetch-space --space 1337 (The space ID)
```

PR #645, opened 2022-03-09
Make menu labels clickable

gustaveen:
## What?
Make menu group labels linkable

## Why?
We have landing pages for Home insurance and should therefore have it's own menu link

## Screenshots / recordings 

https://user-images.githubusercontent.com/6661511/157471425-e6167988-ad5c-42e1-bdb8-71dc113980df.mov

### DIFF
--- a/src/blocks/HeaderBlock/MenuItem.tsx
+++ b/src/blocks/HeaderBlock/MenuItem.tsx
@@ -87,13 +87,14 @@ const GroupLabelListItem = styled(MenuListItem)({
   fontSize: '0.75rem',
 })
 
-const MenuGroupLabel = styled('span')({
+const MenuGroupLabel = styled('a')<{ as?: 'a' | 'span' }>({
   display: 'block',
   color: colorsV3.gray700,
   textTransform: 'uppercase',
   lineHeight: '1rem',
   whiteSpace: 'nowrap',
   paddingBottom: '1rem',
+  textDecoration: 'none',
 
   [TABLET_BP_DOWN]: {
     display: 'inline-block',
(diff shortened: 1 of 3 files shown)


src/blocks/HeaderBlock/MenuItem.tsx
```diff
@@ -273,14 +274,22 @@ export const MenuItem: React.FunctionComponent<{ menuItem: MenuItemType }> = ({
...
+                    {menu_item_groups.map(
+                      ({ label, link, menu_items, _uid }) => (
+                        <DropdownMenuItemList key={_uid}>
+                          <GroupLabelListItem>
+                            {link?.cached_url ? (
+                              <MenuGroupLabel href={getStoryblokLinkUrl(link)}>
+                                {label}
+                              </MenuGroupLabel>
+                            ) : (
+                              <MenuGroupLabel as="span">{label}</MenuGroupLabel>
```
guilhermespopolin, line 286: Is this being rendered as a `span`? I mean, if I understood correctly, `MenuGroupLabel` accepts a prop called `as` which can be `undefined | 'span'` but it does nothing with it, right?

gustaveen: Ah yes! The `as` prop changes the rendered tag, it's a prop for styled components. https://emotion.sh/docs/styled#as-prop But I needed to add the type for it.
The alternative is creating a new style component like so:
```
const MenuGroupLabel = styled(MenuGroupLink)({ cursor: 'default' }).withComponent(
  'span',
)
```

robinandeer (approved):

src/blocks/HeaderBlock/MenuItem.tsx
```diff
@@ -87,13 +87,14 @@ const GroupLabelListItem = styled(MenuListItem)({
   fontSize: '0.75rem',
 })
 
-const MenuGroupLabel = styled('span')({
+const MenuGroupLabel = styled('a')<{ as?: 'span' }>({
```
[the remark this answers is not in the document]
gustaveen: Typescript complained if I didn't have it, and it felt like a good thing to limit it to only a in this case as well `span` :)

PR #648, opened 2022-03-15
Fix/header menu link color

gustaveen:
## What?
- Fix the bug where links in header menu didn't change color on scroll in Chrome
- Fix logo height bug in Safari desktop

## Why?
The bug is that Chrome don't respect the `color: inherit` rule for visited links. It doesn't help if you explicitly set the color on `a:visited` links to inherit. If you visit the site in incognito mode or for the first time the links will update as they should, that's the reason for inconsistent behaviour between users.

So the solution is to set the color explicitly on these links.

## Before

https://user-images.githubusercontent.com/6661511/158408777-bf0a2c0d-95f6-4d3f-bcc7-4d89d37e4a49.mov

## After
https://user-images.githubusercontent.com/6661511/158409374-51092bb4-83dd-4358-9946-f9354f28ebe8.mov

### DIFF
--- a/src/blocks/HeaderBlock/MenuItem.tsx
+++ b/src/blocks/HeaderBlock/MenuItem.tsx
@@ -67,21 +67,24 @@ const DropdownMenuItemList = styled('ul')({
   },
 })
 
-const MenuLink = styled('a')({
-  color: 'inherit',
-  textDecoration: 'none',
+const MenuLink = styled('a')<{ inverseColor?: boolean }>(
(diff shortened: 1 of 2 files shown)


guilhermespopolin (approved):
🤔 that's so weird. I mean, `color: inherit;` should would as expected on this case.
Maybe it's a bug on chrome 🤷🏻

gustaveen:
So weird @guilhermespopolin it should just work. Can't find the page where I read about the issue but it was a chrome bug.

PR #650, opened 2022-03-30
Add hedvig partnership menu

gustaveen:
## What?
Add the possibility to show a different main menu on specific pages.

- Add a new menu, `hedvig_business_menu`, in the global post type
- Add a checkbox (`use_business_menu`) to Header Block to use the Hedvig business menu instead

## Why?
The partnership team wants to have their own top navigation on hedvig.com, Hedvig Business.

We (me, Petar & Mange) discussed if we would need support for more menus and therefore a more dynamic way of adding menus. I looked into possible solutions and we would need to rethink the global post type structure to make that work or build our own plugin. So we decided to make a simple solution for now and iterate on this in racoon instead if we see the need for it.

https://user-images.githubusercontent.com/6661511/160784935-4e9ac2b6-479b-4b95-9a68-990aa3f578aa.mov

### DIFF
--- a/src/blocks/HeaderBlock/index.tsx
+++ b/src/blocks/HeaderBlock/index.tsx
@@ -222,6 +222,7 @@ interface HeaderBlockProps extends BaseBlockPropsDeprecated {
   override_mobile_header_cta_link?: LinkComponent | null
   mobile_header_cta_color?: MinimalColorComponent
   mobile_header_cta_style?: ButtonStyleType
+  use_business_menu?: boolean
 }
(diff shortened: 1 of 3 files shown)


guilhermespopolin (approved):
LGTM
I'm mean, I'm not so familiar with `market-web` but I could understand the reasoning behind the solution and how it was implemented (in essence)
I didn't test it tho 😬

gustaveen:
Thanks @guilhermespopolin 🙏

PR #659, opened 2022-05-03
Feat / Refactor perils and add tab navigation

gustaveen:
## What?
Add tab navigation for Perils instead of current dropdown navigation.
I moved over the Tabs component from web-onboarding and had to do a bit of tweaking to make it fit.

- Refactor `Perils` component to simply accept a collection of perils
- Update the data structure to include an id and label for the tabs. I called it `PerilsCollection`. I therefore renamed some of the components to make more sense.
- Add framer-motion dependency. Our setup doesn't support versions after `4.1.17` (issue described here: https://github.com/framer/motion/issues/1307) and that's the same version we use in `onboarding`. That did cause another issue as well where `tslib` complained `WARNING in ./node_modules/popmotion/dist/es/utils/mix-complex.js 22:15-28
"export '__spreadArray' was not found in 'tslib'` so we had to add that as a dependency as well. More on that issue here: 
https://github.com/framer/motion/issues/1123

Left to do:
- [ ] Update `usePerils` to fetch multiple peril types

## Why?
Consistent UI with Perils in onboarding

## Screenshots / recordings 

https://user-images.githubusercontent.com/6661511/166465273-98b13ac3-9a4d-4174-862e-a3bbb04e3b84.mov


simonauner (approved):
Nice!

I realized after adding comments that you probably just copy-pasted from web-onboarding without any regards to improving the code, in which case you can just disregard my suggestions 😄

gustaveen:
> Nice!
> 
> I realized after adding comments that you probably just copy-pasted from web-onboarding without any regards to improving the code, in which case you can just disregard my suggestions 😄

No that's great! It was just me who missed it so thanks for pointing it out 🙏

src/components/Perils/data/usePerils.ts
```diff
@@ -17,50 +17,62 @@ const getGiraffeEndpoint = (): string => {
...
+
+          perilsArray.push(perilsRequest.data.data.contractPerils)
+        }),
+      )
+      setPerils(perilsArray)
     }
 
     fetchPerils()
-  }, [insuranceType, localeIso])
+  }, [])
```
simonauner, line 75: Lint warning here 👇

simonauner (approved):
Nice! A lint warning that could be nice to get rid of, otherwise LGTM!

src/components/Perils/data/usePerils.ts
gustaveen: Ag yes, I had to memoize the `insuranceTypes` passed to the hook to avoid infinite re-render 😁

simonauner (approved):

src/components/Perils/PerilList/PerilList.tsx
```diff
@@ -0,0 +1,63 @@
+import styled from '@emotion/styled'
+import React from 'react'
+import { LAPTOP_BP_UP, TABLET_BP_UP } from 'components/blockHelpers'
+import { PerilItem } from 'components/Perils/PerilItem/PerilItem'
+import { minimalColorComponentColors } from 'src/blocks/BaseBlockProps'
```
robinandeer, line 5: nit: should this import be updated?

```suggestion
import { minimalColorComponentColors } from 'blocks/BaseBlockProps'
```

src/components/Perils/PerilList/PerilList.tsx
```diff
@@ -0,0 +1,63 @@
...
+  setCurrentPeril: (index: number) => void
+  setIsShowingPeril: (isShowingPeril: boolean) => void
+}
+
+const Wrapper = styled.div`
+  width: 100%;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  margin-left: -0.5rem;
```
robinandeer, line 20: nit: consider CSS grid approach

simonauner (dismissed):

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { createRef } from 'react'
+import styled from '@emotion/styled'
+import { colorsV3 } from '@hedviginsurance/brand'
+import { LAPTOP_BP_UP } from '../blockHelpers'
+import { UnderlineComponent, UNDERLINE_HEIGHT } from './Underline'
+
+const TabContainer = styled.button<{ selected?: boolean }>`
```
robinandeer, line 7: Do we override the focus state globally? Don't remember, otherwise would be nice to add one here :)

src/components/Tabs/Underline.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { motion } from 'framer-motion'
+import { colorsV3 } from '@hedviginsurance/brand'
+// import { useMediaQuery } from 'react-responsive'
+// import { BREAKPOINTS } from 'utils/mediaQueries'
```
robinandeer, line 5: why keep?

robinandeer (dismissed):
Nice work! This is almost nostalgic - I remember how hard/long I worked on this tab solution to make it accessible and get the damn line to animate between the tabs 😅

src/components/Tabs/Tab.tsx
gustaveen: You mean adding a focus state for the tabs? @robinandeer

src/components/Tabs/Underline.tsx
gustaveen: good catch! 🙏

src/components/Perils/PerilList/PerilList.tsx
gustaveen: yup, let's refactor in racoon :)

gustaveen:
Thanks haha but you did all the heavy lifting for this one :D A tricky thing indeed with the line animation! 🙏 @robinandeer

PR #660, opened 2022-05-04
Chore / Remove Codefresh pipeline

robinandeer:
## What?

Remove Codefresh in favor of GH Actions.

@gustaveen we decided that we don't need to include the build step in the CI pipeline, right?

## Why?

Moving away from Codefresh.

**Ticket(s): []**


gustaveen (approved):
Awesome! 🧹

PR #661, opened 2022-05-05
Bugfix / Fix shrinking Peril icon

gustaveen:
## What?
Fix shrinking Peril icon

## Why?

Uploading Screen Recording 2022-05-05 at 11.51.18.mov…


simonauner (approved):
New day, new CSS Flexbox property I haven't seen before :D 

Nice! 🙌

PR #663, opened 2022-05-09
Don't render tabs for single set of Perils

gustaveen:
## What?
Don't render tabs for single set of Perils

## Why?
No need for tab navigation for a single set :) 

https://user-images.githubusercontent.com/6661511/167378117-e5e5f42c-2aba-4ea4-9cac-294d6ac6fa60.mov

### DIFF
--- a/src/components/Perils/index.tsx
+++ b/src/components/Perils/index.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react'
+import React from 'react'
 import { minimalColorComponentColors } from 'src/blocks/BaseBlockProps'
 import { GlobalStory } from 'storyblok/StoryContainer'
 import { Tabs } from '../Tabs/Tabs'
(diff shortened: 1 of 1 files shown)


simonauner (approved):
Sounds reasonable but we do actually render it when only one tab in web-onboarding: https://www.dev.hedvigit.com/se-en/new-member/offer/7d598b69-bc7c-4e3c-b90e-f7512c7a015a?type=SWEDISH_APARTMENT

gustaveen:
> Sounds reasonable but we do actually render it when only one tab in web-onboarding: https://www.dev.hedvigit.com/se-en/new-member/offer/7d598b69-bc7c-4e3c-b90e-f7512c7a015a?type=SWEDISH_APARTMENT

True that! In market-web there is a title above so it makes it a bit overkill :) 
Screenshot 2022-05-09 at 11 23 24

PR #664, opened 2022-05-09
Fix / Alignment peril icons

gustaveen:
## What?
Fix alignment for peril icons

## Why?
The had a transform to offset it slightly. This was actually from a legacy design so not applicable anymore.

https://user-images.githubusercontent.com/6661511/167417046-5410ed6c-bdff-487f-806a-19bac34e0aaf.mov


simonauner (approved):
Nice! 🧹

PR #665, opened 2022-05-10
Make sure perilsArray is returned in correct order

gustaveen:
## What?
Make sure perilsArray is returned in correct order. 

## Why?
Perils showed up in wrong order in the tabs between page loads

### DIFF
--- a/src/components/Perils/data/usePerils.ts
+++ b/src/components/Perils/data/usePerils.ts
@@ -44,9 +44,8 @@ export const usePerils = (
   useEffect(() => {
     const fetchPerils = async () => {
       const url = getGiraffeEndpoint()
-      const perilsArray: Peril[][] = []
 
-      await Promise.all(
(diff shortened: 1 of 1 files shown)


src/components/Perils/data/usePerils.ts
```diff
@@ -44,9 +44,9 @@ export const usePerils = (
   useEffect(() => {
     const fetchPerils = async () => {
       const url = getGiraffeEndpoint()
-      const perilsArray: Peril[][] = []
+      let perilsArray: Peril[][] = []
 
-      await Promise.all(
+      perilsArray = await Promise.all(
```
simonauner, line 49: Can be 
```suggestion
      const perilsArray = await Promise.all(
```
, right?

gustaveen: For sure! 👍

robinandeer (approved):
Nice

PR #666, opened 2022-05-10
GRW-1063 / Feat / Disabled perils

gustaveen:
## What?
Add disabled Perils styling

## Why?
Highlight the difference between insurances

I will talk to Mange to see how we make the icons work. Right now they are exported a bit different so we can't control the fill properly.

What's pending:
- Icons will be updated so we can just change to `fill: currentColor`
- Adam is working on adding a checkbox to PCMS to mark a Perils as disabled 

https://user-images.githubusercontent.com/6661511/167842984-199013e6-7e35-4ff6-8b87-e42df7baa8e7.mov

### Design from Figma

https://www.figma.com/file/kuVQmKybXx4xridPkzVLnD/Car-Insurance-roll-out?node-id=650%3A9178

### DIFF
--- a/src/components/Perils/PerilCollection/PerilCollection.tsx
+++ b/src/components/Perils/PerilCollection/PerilCollection.tsx
@@ -14,6 +14,7 @@ type Props = {
 export const PerilCollection: React.FC<Props> = ({ perils, story, color }) => {
   const [isShowingPeril, setIsShowingPeril] = useState(false)
   const [currentPeril, setCurrentPeril] = useState(0)
+  const enabledPerils = perils.filter((peril) => !peril.disabled)
   return (
     <>
       <PerilList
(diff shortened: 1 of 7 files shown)


src/components/Perils/PerilCollection/PerilCollection.tsx
```diff
@@ -24,7 +25,7 @@ export const PerilCollection: React.FC<Props> = ({ perils, story, color }) => {
       />
       {perils.length > 0 && (
```
simonauner, line 26: `enabledPerils.length` would be more correct, but I guess that it would be a corner case if we have `perils` but no `enabledPerils` 😄

simonauner (approved):
Nice! 

Will the `fill` be added later then when the icons have been updated?

src/components/Perils/PerilCollection/PerilCollection.tsx
gustaveen: Ah yes of course! 🙏

gustaveen:
> Nice!
> 
> Will the `fill` be added later then when the icons have been updated?

🙏 Yes exactly! @simonauner

simonauner (approved):

PR #669, opened 2022-05-13
GRW-1085 / Feat / add TableBlock

gustaveen:
## What?
Add TableBlock to Storyblok. Storyblok has a pretty neat table component we can use. To solve the true/false values in order to show the available/not available icons I went for using `0` or `1`. We can of course extend the use if we want to use the table with text values, but that was nothing that Peter and Mange needed.

## Why?
Possibility to add compare tables inline on the page

Next step is to add a Modal component for this use case. Will gather some feedback from Mange and Peter after trying out this in staging.

https://user-images.githubusercontent.com/6661511/168300150-9c86d1e8-9aa5-4ef9-bf83-86167cfeaac4.mov

### DIFF
--- a/src/blocks/TableBlock/TableBlock.tsx
+++ b/src/blocks/TableBlock/TableBlock.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { colorsV3 } from '@hedviginsurance/brand'
+import styled from '@emotion/styled'
 import {
   Table,
   TableBody,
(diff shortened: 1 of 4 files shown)


src/blocks/TableBlock/TableBlock.tsx
```diff
@@ -0,0 +1,115 @@
...
+              {title}
+            </Heading>
+          )}
+          {text && <Text>{text}</Text>}
+        </TableHeader>
+        <Table fullWidth>
+          {table.thead && (
+            <TableHead>
+              <TableRow>
+                {table.thead.map((cell: any, cellIndex: number) => (
```
simonauner, line 84: You have defined `type TableCell` above, can't that be used here instead of `any`?

src/blocks/TableBlock/TableBlock.tsx
```diff
@@ -0,0 +1,115 @@
...
+                  >
+                    {cell.value}
+                  </TableCell>
+                ))}
+              </TableRow>
+            </TableHead>
+          )}
+          {table.tbody && (
+            <TableBody>
+              {table.tbody.map((row: any) => (
```
simonauner, line 97: same here

simonauner (approved):
Good job! 

Just a question regarding the types of cells, otherwise two thumbs up!

I'm gonna copy the `Dash` icon to market-web also 🎉

robinandeer (approved):
Nice!

Just have the same question about avoiding use of "any" :)

src/blocks/TableBlock/TableBlock.tsx
gustaveen: Of course, late friday brain :D Thank you!

simonauner (approved):

PR #670, opened 2022-05-17
GRW-1091 / Feat / Add ModalBlock

gustaveen:
## What?
Add a new ModalBlock. The use case is when you want to add a modal triggered by a text button. The modal in turn takes other blocks as child blocks. To start with, this is limited to the TableBlock but we can just whitelist more blocks as we see fit.

I did some refactoring on the Modal and added a `maxWidth` prop. 

## Why?
Show table content in a modal.

https://user-images.githubusercontent.com/6661511/168782982-b3bb4ea4-578f-4512-8dbb-d3b048949864.mov

### Editor view
https://user-images.githubusercontent.com/6661511/168783054-75e70942-01d7-4109-a2d3-e8cca1a4bfdc.mov


src/components/Modal/Modal.tsx
```diff
@@ -67,18 +58,21 @@ const ModalContainer = styled('div')<ModalContainerProps>`
       ? 'translateX(-50%) translateY(-50%) scale(1)'
       : 'translateX(-50%) translateY(30%) scale(0.9)'};
 
-  @media (max-height: 900px) {
-    max-height: calc(100vh - 2rem);
+  @media (min-width: 600px) {
```
simonauner, line 61: Isn't it quite a narrow gap here between 600px and (what I assume TABLET_BP_UP is) 768px? Is it needed?

simonauner (approved):
Nice! Just a question about a breakpoint, but it's more of a question that a comment to act on!

src/components/Modal/Modal.tsx
gustaveen: Good point! In this case `TABLET_BP_UP` is 801px (yep, breakpoints are a bit iffy :D ). But let me check if I can change that breakpoint, 768px makes more sense 👍

PR #672, opened 2022-05-23
GRW-1135 / Add compare Perils feature in Storyblok

gustaveen:
## What?
Add possibility to show the differences between set of Perils. It accumulates the different Perils available in the set and shows the difference by a "disabled" look. 

I added a checkbox in Storyblok to enable the feature, will be used on the Car page.

Took inspo from @simonauner's comparison table implementation 😀 

## Why?
The initial plan was to add it as a field in PCMS, but since the release is approaching and it's uncertain if we can expect the feature to be implemented in PCMS before launch we could solve it on our side.

https://user-images.githubusercontent.com/6661511/169812100-0991b032-61ee-4de8-b7a6-3106a841bef6.mov

### DIFF
--- a/src/blocks/PerilsBlock/PerilsBlock.tsx
+++ b/src/blocks/PerilsBlock/PerilsBlock.tsx
@@ -43,10 +43,13 @@ export const PerilsBlock: React.FC<PerilsBlockProps> = ({
     [insurance_types],
   )
 
-  let perils = usePerils(insuranceTypes, currentLocale.iso)
-  perils = compare_perils ? getPerilsComparison(perils) : perils
+  const perils = usePerils(insuranceTypes, currentLocale.iso)
+  const perilsGroup = useMemo(
+    () => (compare_perils ? getPerilsComparison(perils) : perils),
+    [compare_perils, perils],
+  )
 
-  const perilsCollections = perils.map(
+  const perilsCollections = perilsGroup.map(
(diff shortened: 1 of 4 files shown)


src/blocks/PerilsBlock/PerilsBlock.tsx
```diff
@@ -33,22 +35,24 @@ export const PerilsBlock: React.FC<PerilsBlockProps> = ({
...
 }) => {
   const { currentLocale } = useLocale()
   const insuranceTypes = useMemo(
     () => insurance_types.map((insuranceType) => insuranceType.value),
     [insurance_types],
   )
 
-  const perils = usePerils(insuranceTypes, currentLocale.iso)
+  let perils = usePerils(insuranceTypes, currentLocale.iso)
+  perils = compare_perils ? getPerilsComparison(perils) : perils
```
simonauner, line 47: Should this perhaps be in a `useMemo`? Otherwise these will be re-computed ever render, right?

simonauner (approved):
Nice! I wonder if we should have the same in web-onboarding 🤔

src/blocks/PerilsBlock/PerilsBlock.tsx
gustaveen: That's probably a good idea! I'm a bit bad at distinguishing when to use them :)

simonauner (approved):

PR #680, opened 2022-09-08
Upgrade to latest Heroku stack (heroku-22)

gustaveen:
## What?
Upgrade to latest Heroku stack (heroku-22)

Testing the new stack in a review app as suggested by Heroku so we can test it properly before merging and upgrading:
https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack#testing-an-app-on-a-new-stack

## Why?

Screenshot 2022-09-08 at 14 37 00

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,5 +1,6 @@
 {
   "scripts": {},
+  "stack": "heroku-22",
   "env": {
     "USE_AUTH": "true",
     "USE_HELMET": "true",


gustaveen:
Looks like it works good in the review app https://market-web-chore-update-8ypxmd.herokuapp.com/se, images won't get loaded correctly in review apps though, so let's merge and test in staging.

PR #687, opened 2022-10-25
GRW-1583 / Feature / AB redirects from home page

alebedev:
## What?

- Add FEATURE_AB_REDIRECTS flag
- Support redirects to new website, set of source and target pages is hardcoded
- Weights and origin are cofigurable

## Why?

Add ability to test A/B redirects

**Ticket(s): [GRW-1584]**

[GRW-1584]: https://hedvig.atlassian.net/browse/GRW-1584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,3 +10,6 @@ APP_ENVIRONMENT=<development|staging|production>
 # Optional: Datadog Real User Monitoring
 DATADOG_APPLICATION_ID=<datadog application id>
 DATADOG_CLIENT_TOKEN=<datadog client token>
+
+# Feature flags
+FEATURE_AB_REDIRECTS=1
--- a/.env.example
+++ b/.env.example
@@ -13,3 +13,7 @@ DATADOG_CLIENT_TOKEN=<datadog client token>
 
 # Feature flags
 FEATURE_AB_REDIRECTS=1
+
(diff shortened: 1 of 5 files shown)


.env.example
```diff
@@ -9,4 +9,11 @@ APP_ENVIRONMENT=<development|staging|production>
 
 # Optional: Datadog Real User Monitoring
 DATADOG_APPLICATION_ID=<datadog application id>
-DATADOG_CLIENT_TOKEN=<datadog client token>
\ No newline at end of file
+DATADOG_CLIENT_TOKEN=<datadog client token>
+
+# Feature flags
+FEATURE_AB_REDIRECTS=1
```
robinandeer, line 15: Do we need any way to enable this market-by-market or just global on/off?

robinandeer (approved):
Not sure exactly how this will work but code-wise it looks good to me :)

.env.example
alebedev: Global on/off is enough since markets will be addressed in redirect URLs

PR #688, opened 2022-11-15
Use v2 of content delivery API

gustaveen:
## What?
Use v2 of content delivery API

## Why?
Start using Storyblok v2 for market-web since there are no breaking changes to the API and we will set v2 as the default view in the CMS


alebedev (dismissed):
Looks good

alebedev (approved):
Let's do it

PR #689, opened 2022-11-25
TEST: Remove fallback georedirect for SE and add Flagpage

gustaveen:
## What?
Don't use /se as fallback redirect
Add sample Flagpage 

## Why?
Bots should be able to crawl the root of hedvig.com and find links to all locales. Now it's redirect to /se which punishes other markets.

Will test this as a review app

**Ticket(s): []**


robinandeer (approved):
Interesting!

guilhermespopolin (approved):

robinandeer:
@gustaveen do you think we should keep this PR around? Perhaps not needed now that we only support SE, right?

PR #690, opened 2022-11-28
Feat / Native lazy loading

gustaveen:
## What?
Use native lazy loading for images

## Why?
- Improve SEO since bots read all images as the `empty.png` placeholder that was being used before
- Don't use a lib for it now when it's having good support

**Ticket(s): []**


alebedev (approved):
Let's do it

PR #695, opened 2022-12-16
Add country code in html lang

gustaveen:
## What?
Add country code in html lang

## Why?
Signal to google that `en` pages are market specific

**Ticket(s): [GRW-1790]**

[GRW-1790]: https://hedvig.atlassian.net/browse/GRW-1790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/l10n/locales.ts
+++ b/src/l10n/locales.ts
@@ -9,7 +9,7 @@ type MarketName =
   | 'Danmark'
   | 'Denmark'
 type LanguageLabel = 'Sv' | 'En' | 'No' | 'Da'
-type HtmlLang = 'sv' | 'en' | 'no' | 'da'
+type HtmlLang = 'sv-SE' | 'en-SE' | 'no-NO' | 'en-NO' | 'da-DK' | 'en-DK'
 type HrefLang = 'sv-se' | 'en-se' | 'no-no' | 'en-no' | 'da-dk' | 'en-dk'
 
 export type LocaleData = {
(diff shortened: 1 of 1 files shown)


src/l10n/locales.ts
```diff
@@ -9,7 +9,7 @@ type MarketName =
   | 'Danmark'
   | 'Denmark'
 type LanguageLabel = 'Sv' | 'En' | 'No' | 'Da'
-type HtmlLang = 'sv' | 'en' | 'no' | 'da'
+type HtmlLang = 'sv-se' | 'en-se' | 'no-no' | 'en-no' | 'da-dk' | 'en-dk'
```
robinandeer, line 12: I'm not sure if it makes a difference but shouldn't country part be upper case?

https://www.techonthenet.com/js/language_tags.php

robinandeer (dismissed):
Nice, long time coming!

Just curious, what's the difference between HTML lang and HREF lang? Do we need both?

src/l10n/locales.ts
gustaveen: Not sure it matters either but let's play it safe :D

gustaveen:
> Nice, long time coming!
> 
> Just curious, what's the difference between HTML lang and HREF lang? Do we need both?

I think it might be good to keep them apart since they might differ. :)

